PR #188, opened 2020-04-29
Fixes TvDB update frequency option and also fixes types

### LINKED ISSUE #187: 'Calendar update frequency' option in TvDB preferences causes 'Error 400 Bad Request'
Setting the `Calendar` setting in `TvDB preferences` gives a `400 Bad Request` error.
The calendar update frequency is an AniDB setting and not a TvDB setting. The TvDB setting is different.
The cause of the error is that the property name sent to the patch function is `AniDB_Calendar_UpdateFrequency` instead of `UpdateFrequency`

Screenshot_20200429_223323
Screenshot_20200429_223530

As you can see clearly, the property name is wrong.
https://github.com/ShokoAnime/ShokoServer-WebUI/blob/bcd163b6fb7264c985444e68d963dbdb8e791e33/src/core/reducers/settings/Server.js#L66-L72

harshithmohan:
Fixes #187

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jmmserver-webui",
-  "version": "0.3.6-dev3",
+  "version": "0.3.6-dev4",
   "private": true,
   "sideEffects": false,
   "engines": {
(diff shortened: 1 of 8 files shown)


harshithmohan:
The property `value` in `SettingsDropdown` had to be set to `PropTypes.any` because it receives both `string` and `number` data types.
`string` for the language selection dropdown and `number` for everything else.

PR #192, opened 2020-05-22
Converted to TS and started new WebUI

harshithmohan:
- **Full conversion to TypeScript (except tests)**
- I've deleted a lot of things (well, most of the things, all related to UI) when converting to TS to get it to a working state. I will put them back one by one and convert them to TS when I get to redesigning them.
- Completely removed `Bulma`
- Partially removed `BlueprintJS`
- For now, only the login page and dashboard (without header) exist.
- Have to fix a problem with fonts. There are multiples of same font files (with different weights) because `font-weight` isn't working properly.
- Have to clean and organize the CSS files and classes.
- Have to cleanup unneeded dependencies.
- Also, haven't written or converted any test files yet.


harshithmohan:
I blanked out the tests because it wouldn't pass as I have not converted the tests to TS.

hidden4003:
Typescirpt has a mode that it can just ingest js files without any typescript stuff in them if you leave js extension, why wipe everything ? I understand typing will have to go but overall it is easier maybe?

harshithmohan:
TS can ingest JS files properly, that's ok. The problem is with Flow.
The TS compiler complains about every non-JS Flow thing like typing, importing types etc.
If Flow wasn't here, I could've just kept both JS and TS files.
So, either I deleted the files or removed the typings/converted to TS to make it even start.
I just deleted the files containing UI because I would rewrite them anyway and I converted all the other things like reducers, actions etc because those would be just re-used.

PR #197, opened 2020-08-02
Poll dashboard data only on general queue change

harshithmohan:


src/core/middlewares/signalr.ts
```diff
@@ -19,7 +19,10 @@ const onQueueStateChange = (dispatch, getState) => (queue, state) => {
   dispatch(setQueueStatus(newState));
 };
 const onQueueCountChange = (dispatch, getState) => (queue, count) => {
-  const newState = Object.assign({}, { [queue]: Object.assign({}, getState().mainpage.queueStatus[queue], { count }) });
+  const currentQueue = getState().mainpage.queueStatus[queue];
+  if (queue === 'general' && currentQueue.count > count) dispatch({ type: Events.MAINPAGE_REFRESH });
```
hidden4003, line 23: It would be good to add throttle around dispatch({ type: Events.MAINPAGE_REFRESH })

PR #235, opened 2021-05-12
[WIP] WebUI v2 changes

harshithmohan:


src/core/sagas/settings.ts
```diff
@@ -56,7 +56,7 @@ function* getPlexLoginUrl() {
 }
 
 function* getSettings() {
-  yield put(startFetching('settings'));
+  // yield put(startFetching('settings'));
```
hidden4003, line 59: should this be commented out ?

src/core/sagas/webui.ts
```diff
@@ -14,7 +14,7 @@ import { setItem } from '../slices/misc';
 
 function* checkUpdates() {
   yield put(startFetching('checkingUpdates'));
-  const channel = yield select((state: RootState) => state.webuiSettings.v3.updateChannel);
+  const channel = yield select((state: RootState) => state.webuiSettings.webui_v2.updateChannel);
```
hidden4003, line 17: do we really want to keep v1 and v2 in same key? Maybe add some parameter to settings for version and server could use totally separate clean key or it ?

src/core/slices/webuiSettings.ts
```diff
@@ -78,34 +80,34 @@ export const defaultLayout = {
...
-      i: 'trakt', x: 8, y: 28, w: 4, h: 5, minW: 3, minH: 3, moved: false, static: false,
+      i: 'trakt', x: 8, y: 39, w: 4, h: 9, minW: 3, minH: 3, moved: false, static: false,
     }],
-  },
+  } as LayoutType,
 };
 
 const initialState = {
-  v3: {
+  webui_v2: {
```
hidden4003, line 110: same with regard to server managing settings versioning, should not be necessary, also maybe just use key name like "server" to indicate it is managed by server. I am thinking of possible clash during migration and what not maybe version number is inside the server response then we can at least put some code like - you are using webui v1, you should use v2 etc. in case server gives different settings object.

src/pages/firstrun/StartServer.tsx
```diff
@@ -1,74 +1,60 @@
...
+
+  const handleNext = () => {
+    dispatch({ type: Events.STOP_API_POLLING, payload: { type: 'server-status' } });
+
+    dispatch({
+      type: Events.AUTH_LOGIN,
+      payload: {
+        user: user.Username,
+        pass: user.Password,
+        device: 'web-ui',
```
hidden4003, line 25: Maybe use different string for v2 ?

src/core/sagas/settings.ts
harshithmohan: Yes, I did some "magic" there.
So that when you save a setting, the spinner doesn't show up.
I'll have to rewrite the code there to be more understandable.

src/pages/firstrun/StartServer.tsx
harshithmohan: Yeah, I could do it. But it shouldn't matter, right?

hidden4003 (approved):

PR #252, opened 2021-10-12
Project improvements

Codex-:
Just a small set of project improvements based on what I saw at first glance when going to implement some features.

- @types are dev deps, not real deps, unless exported explicitly as a part of a package to be consumed.
- Updated the gitignore with things that were immediately generated and then git tracked.
- Abstracted the eslint rules out to a dedicated `.eslintrc.json`
  - Sorted / organised the file.
  - Added section to handle the JS config explicitly.
    - Means you can avoid the copy paste ignores.
  - Added `@typescript-eslint/await-thenable` & `@typescript-eslint/no-floating-promises`.
    - These two warn on what in my experience is one of the most common sources for bugs in applications.
- Improve the CICD workflows.
  - Bump to use latest `setup-node`.
  - Run lint / build and every commit.
  - Only generate artifacts for PR's (original behaviour).
- Update use of deprecated webpack config.


hidden4003:
master is protected branch so you can't push to it, which is why tests are only on PR

Codex-:
> master is protected branch so you can't push to it, which is why tests are only on PR

Yeah, of course. I'm not sure I understand your point here.

If you mean the workflow change on push, then this change allows tests to run per commit irrespective of the branch. Extremely useful when doing a feature branch before you get to the PR stage.

PR #253, opened 2021-10-15
Project improvements

Codex-:
Just a small set of project improvements based on what I saw at first glance when going to implement some features.

- @types are dev deps, not real deps, unless exported explicitly as a part of a package to be consumed.
- Updated the gitignore with things that were immediately generated and then git tracked.
- Abstracted the eslint rules out to a dedicated `.eslintrc.json`
  - Sorted / organised the file.
  - Added section to handle the JS config explicitly.
    - Means you can avoid the copy paste ignores.
  - Added `@typescript-eslint/await-thenable` & `@typescript-eslint/no-floating-promises`.
    - These two warn on what in my experience is one of the most common sources for bugs in applications.
- Improve the CICD workflows.
  - Bump to use latest `setup-node`.
  - Run lint / build and every commit.
  - Only generate artifacts for PR's (original behaviour).
- Update use of deprecated webpack config.


Codex-:
Rebased on webui v2 branch.

src/core/middlewares/signalr.ts
```diff
@@ -121,8 +121,8 @@ const signalRMiddleware = ({
     connectionLog.on('Log', onLogsLog(dispatch));
     startSignalRConnection(connectionLog);
   } else if (action.type === Events.AUTH_LOGOUT) {
-    if (connectionEvents !== undefined) connectionEvents.stop();
-    if (connectionLog !== undefined) connectionLog.stop();
+    await connectionEvents?.stop();
+    await connectionLog?.stop();
```
Codex-, line 125: If you want to float these promises then consider using an IIFE to make it explicitly clear that a floated promise is the intention, that is the intention behind how the eslint rule is configured.

hidden4003 (approved):

.github/workflows/CI.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         with:
           node-version: 14
       - name: Install dependencies
-        run: yarn install --frozen-lockfile
+        run: yarn install --immutable-cache # lock changes should already be committed.
```
hidden4003, line 18: docs say you are supposed to use --immutable as a replacement for --frozen-lockfile, not --immutable-cache

.github/workflows/Release.yml
```diff
@@ -12,11 +12,11 @@ jobs:
...
         uses: actions/checkout@v2
       - name: Setup Node.js v14
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: 14
       - name: Install dependencies
-        run: yarn install --production=false
+        run: yarn install --frozen-lockfile --production=false
```
hidden4003, line 19: needs to be --immutable here as well

hidden4003 (changes requested):

.github/workflows/CI.yml
Codex-: `immutable` is default enabled in CI environments:
```
If the --immutable option is set (defaults to true on CI)
```

https://yarnpkg.com/cli/install
image

`--immutable-cache` is more desirable behaviour for ensuring that CICD is running the intended and committed state of the repository.

.github/workflows/Release.yml
Codex-: Ah damn, will update this one. Probably shouldn't have the `production` modifier either

.github/workflows/CI.yml
Codex-: Hmm odd, removed for now.

PR #312, opened 2022-07-13
Fix episode ids on daily and more

revam:
Fix episode ids on daily and add duration, resume position and watched date

### DIFF
--- a/src/core/types/api/dashboard.ts
+++ b/src/core/types/api/dashboard.ts
@@ -26,13 +26,20 @@ export type DashboardStatsType = {
 };
 
 export type DashboardEpisodeDetailsType = {
-  ID: number,
+  IDs: {
+    ID: number,
+    Series: number,
+    ShokoFile: number | null,
+    ShokoEpisode: number | null,
+    ShokoSeries: number | null,
+  },
   Title: string,
   Number: number,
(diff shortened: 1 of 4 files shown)


src/core/types/api/dashboard.ts
```diff
@@ -26,13 +26,20 @@ export type DashboardStatsType = {
 };
 
 export type DashboardEpisodeDetailsType = {
-  ID: number,
+  IDs: {
+    ID: number,
+    Series: number | null,
```
hidden4003, line 31: we use Series?: number; instead of | null;

hidden4003 (changes requested):

src/core/types/api/dashboard.ts
revam: But it will be a number or null, not a number or undefined. Doing a(n accidental) strict equal to undefined would result in false if the value is null.
Also, the (anidb) series id will always be set and never be null. Sorry,  my (1am) mistake, will fix it now.

hidden4003: I thought it was null, but if it is undefined then leave it as is.

hidden4003 (approved):

PR #314, opened 2022-07-14
Add recently added series and more

revam:
### Changes

- Add recently added series tab to the recently imported tab on the dashboard

- Added all the series definitions and some common definitions to support the series definitions.

- Changed existing definitions to use the new common types (to reduce code duplication for the same types used in different places).

- Fixed the date and time display for the calendar.

- Fixed the initial state for the calendar to align with the state elsewhere.

### DIFF
--- a/src/pages/dashboard/panels/UpcomingAnime.tsx
+++ b/src/pages/dashboard/panels/UpcomingAnime.tsx
@@ -17,7 +17,7 @@ const Title = ({ showAll, setShowAll }) => (<div>
 
 const UpcomingAnime = () => {
   const items = useSelector((state: RootState) => state.mainpage.upcomingAnime);
-  const [showAll, setShowAll] = useState(true);
+  const [showAll, setShowAll] = useState(false);
 
   const dispatch = useDispatch();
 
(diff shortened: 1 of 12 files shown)


src/pages/dashboard/panels/UpcomingAnime.tsx
```diff
@@ -28,10 +28,15 @@ const UpcomingAnime = () => {
 
   const renderDetails = (item: DashboardEpisodeDetailsType ) => {
     const airDate = moment(item.AirDate);
-    
+    let relativeTime = airDate.fromNow();
+    if (relativeTime.includes('ago'))
+      relativeTime = 'today';
```
hidden4003, line 33: for this you want to use moment.relativeTimeThreshold(unit, limit) to change the threshold where is starts using hours instead days etc.. and you can set it to use 'today' for anything below 24h too.

hidden4003 (changes requested):

src/pages/dashboard/panels/UpcomingAnime.tsx
revam: I can't seem to find a good alternative way using the moment.js library to get it to display "today" for the current date instead of counting in hours/minutes/seconds ago. Changing the thresholds doesn't seem to be the correct answer and their doc site isn't much help...

Also, I forgot to address the case for "tomorrow" if you're minutes or seconds from midnight.

hidden4003 (approved):

PR #334, opened 2022-07-17
Update first run and login pages

harshithmohan:


src/pages/firstrun/Acknowledgement.tsx
```diff
@@ -19,27 +19,27 @@ function Acknowledgement() {
 
   return (
     <TransitionDiv className="flex flex-col text-justify justify-center px-96">
-      <div className="font-semibold text-lg">Thanks For Installing Shoko Server!</div>
-      <div className="font-open-sans font-semibold mt-10">
+      <div className="font-semibold">Thanks For Installing Shoko Server!</div>
+      <div className="mt-9">
```
hidden4003, line 23: you can put space-y-9 on parent instead of putting mt-9 on every item

hidden4003 (approved):

PR #335, opened 2022-07-17
Refactor dashboard panels to reuse components.

revam:


src/pages/dashboard/components/DashboardTitleToggle.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import cx from 'classnames';
+
+const DashboardTitleToggle = ({ mainTitle, secondaryTitle, secondaryActive, setSecondaryActive }: { mainTitle: string; secondaryTitle: string; secondaryActive: boolean; setSecondaryActive(value: boolean): void }) => (<div>
```
hidden4003, line 4: ShokoPanelTabsTitle would probably make more sense as other pages have it too, like utilities, but we can rename later

hidden4003 (approved):

PR #348, opened 2022-07-23
Fix api

revam:
Blocked by and dependent on https://github.com/ShokoAnime/ShokoServer/pull/947

### Changes

- Remove the unused code for getting the details of the recent files

- Fix the bits that broke by https://github.com/ShokoAnime/ShokoServer/pull/947

- ~~Pages start at `0`.~~

- Added the AniDB file type, exported the file ids type and adjusted the dashboard stats and summery types


revam:
If the page start at `1` then it needs to be subtracted by `1` before it's send through the api

revam:
> If the page start at `1` then it needs to be subtracted by `1` before it's send through the api

It's now offset by `-1` before it's sent through the api.

hidden4003 (approved):

PR #365, opened 2022-07-31
Change group page load

revam:


src/pages/collection/CollectionPage.tsx
```diff
@@ -123,19 +122,17 @@ function CollectionPage() {
...
-        return (
-          <div key={key} style={style}>
-            {renderPlaceholder()}
-          </div>
-        );
-      }
-      return null;
+    const neededPage = Math.ceil((index + 1) / pageSize);
+    const groupList = fetchedPages[neededPage];
+    if (groupList == undefined) {
```
revam, line 127: Before somebody asks, this is on purpose so it will check for both `undefined` and `null`.

hidden4003 (approved):

PR #409, opened 2022-09-29
Information panels for Actions and Utilities

LexMalta:
As a new user of Shoko I have no idea what all the different actions and utilities do - so I wanted to add a description of each of them into the UI.

@ElementalCrisis put together a design on discord which I'm intending to follow

I've created info content based on docs in Swagger and my understanding of how Shoko works, so if I've got something wrong please correct me! I've put stars on the descriptions I think could be improved.


harshithmohan:
Not against writing it in scss but it would be better if there's a way to do the animations with just tailwind classes

LexMalta:
@harshithmohan I had a try at doing it that way, but I don't think it's possible to both define the keyframes and then use them for an animation all in one go?

harshithmohan:
We could go for a simpler animation maybe 🤔 
So, click and it just expands from the top, so a transition on height change perhaps. (that wouldn't need keyframes, I'm thinking)
That would match the style of other animations
@ElementalCrisis what do you think?

harshithmohan:
Also @LexMalta, you might want to check the linter errors

ElementalCrisis:
Make the tooltip width fixed as we don't want the sidepanel's total width to expand on click. As for the animation, I'd prefer a smoother one, maybe the same one we use for the headers themself?

src/components/Dialogs/ActionsModal.tsx
```diff
@@ -77,6 +77,30 @@ const actions = {
...
+const RenderItem = (item: { name:string, function:string, data?:boolean, runAction:Function }) => {
+  const [showInfo, setShowInfo] = useState<boolean>(false);
+
+  return (
+    <>
+      <div className="flex justify-between font-semibold">
+        <span>{item.name}</span>
+        <div>
+          <span className="inline-block pr-2 text-highlight-1" onClick={() => setShowInfo(!showInfo)}><Icon className="cursor-pointer" path={mdiInformationOutline} size={1}></Icon></span>
+          <span className="inline-block text-highlight-1" onClick={() => item.runAction(item.name, item.function)}><Icon className="cursor-pointer" path={mdiPlayCircleOutline} size={1} /></span>
```
revam, line 89: Why not just pass a `onClick` prop instead of the `function` and `runAction` props?

src/components/Dialogs/ActionsModal.tsx
```diff
@@ -93,21 +117,14 @@ function ActionsModal() {
...
-
   const renderTab = (title, items) => (
     <React.Fragment>
       <div className="grow px-4 py-2 bg-background-alt self-stretch border-b border-background-border shadow flex justify-between">
         <p className="text-base font-semibold text-gray-300">{title}</p>
         <span onClick={() => { setActiveTab(title); }}><Icon className="cursor-pointer" path={mdiChevronUp} size={1} rotate={activeTab === title ? 0 : 180} /></span>
       </div>
       <div className={cx('flex flex-col grow w-full p-4 space-y-1', { hidden: activeTab !== title })}>
-        {items.map(item => renderItem(quickActions[item]))}
+        {items.map(item => <RenderItem {...quickActions[item]} runAction={runAction} />)}
```
revam, line 127: If we can avoid it then please don't use the spread operator when rendering react elements. Also, if we don't spread then we don't need to pass the unused `data` property (until it's used at least) amd can omitn8t from the props object used by the `RenderItem` function.

src/components/Dialogs/ActionsModal.tsx
```diff
@@ -12,6 +12,7 @@ import { setStatus } from '../../core/slices/modals/actions';
 import { useRunActionMutation } from '../../core/rtkQuery/actionsApi';
 
 import { RootState } from '../../core/store';
+import { Transition } from '@headlessui/react';
```
revam, line 16: Linter error for unused import.

src/components/Dialogs/ActionsModal.tsx
```diff
@@ -77,6 +78,30 @@ const actions = {
   },
 };
 
+const RenderItem = (item: { name:string, onClick:Function, info?:string }) => {
```
revam, line 81: Much better 😃👍

src/core/quick-actions.ts
```diff
@@ -75,11 +75,13 @@ const quickActions = {
     name: 'Remove Missing Files',
     function: 'RemoveMissingFiles',
     data: true,
+    info: 'Removes entries in Shoko and MyList for files thats are no longer accessible',
```
revam, line 78: If you want then i can help adding the descriptions.

revam (changes requested):
Please fix the linter error. 😅

Otherwise it is looking good so far. 😃👍

revam:
Also, some small advice; I will recommend running `yarn run lint` before pushing to catch any lint errors, because otherwise will the CI pipeline complain if you push and the changes contain any lint errors. 😅

ElementalCrisis:
Looking good @LexMalta, resolve the lint error and we can get this merged in.

hidden4003 (approved):

PR #410, opened 2022-10-02
Fix crash in UpcomingAnime

LexMalta:
I'm running the latest stable version of Docker Shoko server

When logging into the WebUI, I get the following crash

image

If I log the `items` in `UpcomingAnime.tsx` I get the following:
image

Which has no `items.data.IDs` property (hence the crash) and looks nothing like a `DashboardEpisodeDetailsType` that `EpisodeDetails` is expecting.

I imagine that this is because I'm running a version of Shoko server that's out of sync from the Web-UI, but if we release the Web-UI and users haven't updated data on their server they're likely to run into this?


ElementalCrisis:
Seems I'm not the only one who gets this error, @hidden4003 and @harshithmohan, lol.

harshithmohan (approved):

hidden4003:
This does not look like a fix it just ignores the bad format. If you want this to be merged, please, just fix the ID to the actual data you receive. Also, @revam, did the format actually change ?

revam:
@LexMalta  Use a daily version of shoko server when testing the new web ui please.

@hidden4003 and yes it did. From swagger;
image

PR #416, opened 2022-10-15
fix: don't assume unrecognised files have a location :adhesive_bandage:

revam:


harshithmohan:
da3d mentioned a file should exist in the DB without a location so this is fixing a problem which is not ours to fix.

bigretromike:
The only event that would result in unrecognized file without location would be non-cascade deletion of that file from db ?
Any other ?

hidden4003 (approved):

PR #454, opened 2022-11-23
Fix log tab streaming

hidden4003:


da3dsoul:
What was the issue? It seemed like it just stopped working, and I don't think I broke it

hidden4003:
> What was the issue? It seemed like it just stopped working, and I don't think I broke it

@da3dsoul  I ported the code wrong when migrating to rtk query and didn't notice because initial load worked but streaming didn't.

da3dsoul:
Ahhh fair enough

PR #457, opened 2022-11-25
update version types and usage

revam:
also, added the commit behind the version when not in the stable,
  and the channel before the version on the first-run and login pages.


revam:
also, the old web ui is **unusable** with this version. we may need a backwards-compatible fix for new v1 installs until v2 is out.

hidden4003 (approved):

PR #485, opened 2022-12-08
Add image placeholder

harshithmohan:


src/components/BackgroundImagePlaceholderDiv.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+function BackgroundImagePlaceholderDiv(props: Props) {
+  const {
+    children, className,
+    imageSrc, hidePlaceholderOnHover,
+  } = props;
+
+  const [imageLoaded, setImageLoaded] = useState(false);
+
+  const backgroundImage = new Image();
```
revam, line 21: We should probably use a ref or memo to avoid creating a new Image object on every render....

hidden4003: either memo this or React.memo whole component

hidden4003 (changes requested):
BackgroundImagePlaceholderDiv needs React.memo

hidden4003 (approved):

PR #494, opened 2023-01-12
Spellcheck

krbrs:


src/core/quick-actions.ts
```diff
@@ -52,7 +52,7 @@ const quickActions = {
   'avdump-mismatched-files': {
     name: 'AVDump Mismatched Files',
     function: 'AVDumpMismatchedFiles',
-    info: 'Used by the \'Unrecognised\' utility to find files in your library that don\'t match to a file in AniDB.',
+    info: 'Used by the \'Unrecognized\' utility to find files in your library that don\'t match to a file in AniDB.',
```
revam, line 55: Tbh i prefer non American spellings

krbrs: But the utility itself is also written with Z 😅 I never know there was a BE variant with "s" at all

harshithmohan: Indian English says "unrecognized" is the correct one.
I have spoken 🙃

revam: 🤷‍♂️

da3dsoul: IMO, the way it's said (American) is the correct spelling. I'm sure there's an accent that refutes that, but meh. Every accent I've heard still uses a Zed

revam: I'll just borrow @krbrs's message from discord and post it here;

> Unrecognized and unrecognised are both English terms. Unrecognized is predominantly used in 🇺🇸 American (US) English ( en-US ) while unrecognised is predominantly used in 🇬🇧 British English (used in UK/AU/NZ) ( en-GB ). In the United States, there is a preference for "unrecognized" over "unrecognised" (98 to 2).

Also, I'll let @hidden4003 decide what to do.

hidden4003: Looks like unrecognized is the majority.

hidden4003 (approved):

PR #496, opened 2023-02-26
Add fallback output for when there are zero manually linked files

fearnlj01:
As I seem to type on all my PRs, there's probably a better way of doing this but making simple small changes is the way I roll!

This change should prevent the `TypeError: virtualItems[0] is undefined` error from appearing on the manually linked files tab when there are no manually linked files in Shoko.

I've tested this on my machine™ with this change, starting from having no manually linked files to then manually linking a file in the unrecognised files tab, unlinking it from the manually linked tab and finally deleting the file back on the unrecognised files tab without any errors being thrown at any point.


hidden4003 (approved):

hidden4003:
This obviously needs a refactor but as a quick fix it is fine.

PR #521, opened 2023-04-09
feat: Add missing quick actions

revam:


revam:
Sync hashes is disabled in the server and only returns "400" (so it's essentially a no-op). It's a remains from the old web cache functionally (that's now de-functional). So, I removed the action for it.

PR #537, opened 2023-05-31
Pulse cell background on jump - Collection

ItzzLincler:
Just a quality of life thing, it was killing me that I didn't know where the shows I was looking for started after I jumped to a certain letter.
Right now it keeps pulsing forever, not sure if I should limit it or leave as is (I kinda prefer it like it is)

### DIFF
--- a/src/pages/collection/GroupList.tsx
+++ b/src/pages/collection/GroupList.tsx
@@ -19,7 +19,7 @@ import { resetGroups, setGroups } from '../../core/slices/collection';
 import { useLazyGetGroupViewQuery } from '../../core/rtkQuery/splitV3Api/webuiApi';
 
 import type { RootState } from '../../core/store';
-import Jumpbar from './Jumpbar';
+import Jumpbar, { HighLightTargetType } from './Jumpbar';
 
 function GroupList() {
   const itemWidth = 240; //224 + 16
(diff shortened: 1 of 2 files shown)


revam:
Also, you should probably rebase this for `design-v7`, otherwise you will probably have to wait for that to be merged back to master (and there will probably be merge conflicts).

revam:
@ElementalCrisis what do you think of the pulsing animation they suggested for the loading?

@hidden4003 can you look over the implementation (potentially after they've switched over the target branch to `design-v7`, rebased their changes, and fixed their linting error)?

hidden4003:
@revam there is no jumpbar in design-v7 branch anymore, so I am unsure how that rebase would go

hidden4003:
@ItzzLincler very hard to review as 90% of the changes are formatting, we have a linter in this project that enforces certain style, your formatting seems to be set to opposite of that, also the jumpbar no longer exists in design-v7 branch, which will be merged soon into master

As for implementation I think better way is to use the indexes from the jumpbar and pass on the start index and length, or end index as for certain symbols like # in the jumpbar there is not a direct 1:1 relation with it being 1st letter of a title, so it will not work on those items.

ElementalCrisis:
For reference. 

https://github.com/ShokoAnime/ShokoServer-WebUI/assets/9443295/adb3b6ea-8cf6-49b5-89d2-27a5345140f2

This is a neat idea; I'll see if there's somewhere in v7 we can use this.

ItzzLincler:
@revam Sorry about the linting mismatch, I was just used to formatting with prettier and didn't even notice It was overriding it.

@hidden4003 @ElementalCrisis Honestly my current solution was only supposed to be a quick stop gap of a solution that could be easily pushed into daily builds.

long term I think a better solution is to have dividers for each category/letter/number/symbol so you could just jump to the top of the proper divider plus you don't need the animation to easily find what you are looking for and it looks cleaner.
Ex:
image

I will take a look at the v7 branch and see if it can be implemented easily. 

FYI: I am kinda new to react/JS so I might write shit code... sorry in advance

ElementalCrisis:
@ItzzLincler with the v7 changes merged in, this PR is unfortunately no longer viable as we no longer have the jump bar. I'm going to close it but I'd love to see more outside perspective on changes. If you're in our Discord, send me a message and I'll add you to the Web UI beta group.

PR #542, opened 2023-06-04
Allow using a non default proxy port

ItzzLincler:
Quality of life when developing, saves me the hassle of changing the port every time I pull and ignoring changes to the run.js file

### DIFF
--- a/run.js
+++ b/run.js
@@ -147,7 +147,7 @@ tasks.set('start', () => {
     const middleware = [];
     if (config.apiProxyIP) {
       const proxyMiddleware = createProxyMiddleware(['/api', '/plex', '/signalr'], {
-        target: `http://${config.apiProxyIP}:8111`,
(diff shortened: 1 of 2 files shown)


ItzzLincler:
*The change will require people to update the config file with their port

run.js
```diff
@@ -145,9 +145,9 @@ tasks.set('start', () => {
     });
 
     const middleware = [];
-    if (config.apiProxyIP) {
+    if (config.apiProxyUrl) {
       const proxyMiddleware = createProxyMiddleware(['/api', '/plex', '/signalr'], {
-        target: `http://${config.apiProxyIP}:8111`,
+        target: `http://${config.apiProxyUrl}`,
```
hidden4003, line 150: Add the protocol to the url as well so it is fully customizable

hidden4003 (changes requested):

hidden4003 (approved):

PR #545, opened 2023-06-04
Rewrite collection with tanstack virtual, move non-page components

harshithmohan:


src/components/Collection/CollectionView.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+          // Here, i will be the actual index of the group in group list
+          for (let i = fromIndex; i < toIndex; i++) {
+            const neededPage = Math.ceil((i + 1) / pageSize);
+            const relativeIndex = i % pageSize;
+            const groupList = groupPages[neededPage];
+
+            if (groupList === undefined && !fetchingPage) {
+              setFetchingPage(true);
+              fetchPage(neededPage);
```
hidden4003, line 121: This does not have to be here, can be moved safely outside the for loop and improve performance and readability a bit, essentially you only need to check 2 indices, the starting one and ending one and in-between you can skip to determine pages you need to load.

src/components/Collection/CollectionView.tsx
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            const item = groupList !== undefined ? groupList[relativeIndex] : undefined;
```
hidden4003, line 124: If you are on last/not full page, doesn't it mean you will start getting empty items here?

hidden4003 (approved):

src/components/Collection/CollectionView.tsx
harshithmohan: `groupList[relativeIndex]` will return `undefined` and then in the next line `isPlaceholder` will become true, which is needed to render empty divs.
Without the placeholder divs, the items in the last row will be center aligned as each row has `justify-center` to not have more empty on the right than the left

src/components/Collection/CollectionView.tsx
harshithmohan: Moved it outside the loop

PR #552, opened 2023-06-05
Update LanguagesModal.tsx

krbrs:
Change "Kanji" to "Japanese (ja)" to be consistent with commons


harshithmohan:
Can you place the `Pinyin` also at the right to not trigger ocd?

src/components/Dialogs/LanguagesModal.tsx
```diff
@@ -10,7 +10,7 @@ import { useGetSettingsQuery, usePatchSettingsMutation } from '@/core/rtkQuery/s
 export const languageDescription = {
   'x-jat': 'Romaji (x-jat)',
   en: 'English (en)',
```
revam, line 12: @krbrs place `Pinyin (x-zht)` above `English (en)`. I'll add it to the default list on the server side too, since donghua have `x-zht` for their main title and @harshithmohan wanted it to be in the default list for it to be placed near `Romaji (x-jat)`.

revam:
Time to update the defaults on the server,  i guess.

PR #555, opened 2023-06-06
Switch from Webpack to Vite

funniray:
Vite vastly simplifies the entire build pipeline of the app, while also improving build times.

On my machine, I was able to see a reduction in build time from 25 seconds, down to just 11 seconds.
Additionally, HMR start times went from 18 seconds, down to just 1 second.

Notable changes:
- Webpack and all loaders/plugins were dropped.
- `run.config.js` was replaced with `proxy.config.js` -- any configuring of the build should be done using a .env file.
- No `vite.config.default.js` was created. I didn't do this as vite lets you specify the config via the cli. 
- `run.js` was removed, in favor of running vite standalone and modifying CI pipelines to zip up the dist folder.
- `sass` was dropped in favor of raw postcss. This is what's recommended by tailwind. No scss features were used.
- css was moved from `/css` to `/src/css`

What I need help/input on:
- browsersync isn't currently setup. There is a [vite plugin](https://www.npmjs.com/package/vite-plugin-browser-sync) for it, but I'm unsure if anybody uses it, or if it should be included by default?
- Are any dependencies still included in the package.json, but are no longer requred?


index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+      <meta charset="utf-8">
+      <meta name="viewport" content="width=device-width, initial-scale=1.0">
+      <title>Shoko WEB UI</title>
+      <link rel="shortcut icon" href="favicon.ico" />
+  </head>
+  <body>
+    <link rel="stylesheet" type="text/css" href="/api/v3/WebUI/Theme.css" /> <!-- This is inside body because it needs to be loaded after the style tags in dev server -->
```
revam, line 10: This should only be in the body for the dev server though.

funniray: I'm not quite sure how Theme.css works, to be completely honest, I was testing on an old build of shoko server

.github/workflows/Release-Dev-Auto.yml
```diff
@@ -83,4 +84,4 @@ jobs:
             Update from the WebUI or click [here](https://github.com/${{ github.repository }}/releases/tag/v${{ steps.semver.outputs.version_tag }}) to download!
             
             **Changelog**:
-            ${{ env.CHANGELOG }}
\ No newline at end of file
+            ${{ env.CHANGELOG }}
```
revam, line 87: why did this change?

funniray: Was an accident, my IDE did it on it's own. I tried to look at most of the diffs

tailwind.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
   ],
   content: [
     './src/**/*.tsx',
-    './public/**/*.html',
+    '.index.html',
```
revam, line 19: typo?

.eslintrc.json
```diff
@@ -30,6 +28,7 @@
     ],
     "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
     "implicit-arrow-linebreak": "off",
+    "import/extensions": "off",
```
harshithmohan, line 31: Why is this turned off?

index.html
revam: The server just provides an auto-generated css file with the theme declarations. The theme declarations are just like the `.theme-shoko-gray` declaration (now) found under `src/css/theme-gray.css` (which should be renamed to `src/css/theme-shoko-gray.css` imo), and applying a theme is as simple as applying the theme class name to an element, be it our root element or any other element on the page.

funniray:
> can you set build.chunkSizeWarningLimit higher so it stops with the warning as we are not going to do dynamic imports

A few things probably could get dynamically imported to reduce part of the base bundle size, like Sentry, which takes up 350kb of the current bundle.

funniray:
Probably not 100% relevant to this PR, but is there any reason why code splitting/dynamic imports shouldn't be used? From what I've seen:

- It allows multiple chunks to be downloaded at once
- Doesn't parse javascript from a chunk until it's required
- Decreases the time until first draw

Loading screens while switching routes can be minimized by prefetching content, while still lazy loading them.

index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Shoko WEB UI</title>
+    <link rel="shortcut icon" href="/favicon.ico" />
```
revam, line 7: `/favicon.ico` doesn't exist, and neither does the Web UI live at the root either. Prefix the icon with `/webui` to make it accessible from any url location within the Web UI.
```suggestion
    <link rel="shortcut icon" href="/webui/favicon.ico" />
```

vite.config.js
```diff
@@ -0,0 +1,62 @@
...
+      sourcemap: true,
+    },
+    plugins: [react(), sentryPlugin],
+    base: "/webui/"
+  };
+});
+
+async function setupEnv(isDebug) {
+  const gitHash = childProcess.execSync("git log --pretty=format:'%h' -n 1").toString().replace(/["']/g, '');
+  const appVersion = JSON.stringify(pkg.version);
```
revam, line 53: Why are we trying to stringify a value that's already a string? I don't think we will ever change it so it's not a string in the `package.json` file, so it's safe to just directly use it as a string here.
```suggestion
 const appVersion = pkg.version;
```

revam (changes requested):
Looked through it. It is mostly looking good now, except the three comments I posted below.

index.html
funniray: the `/webui/ ` prefix is automatically added by Vite with the `base` option.

vite.config.js
funniray: was a relic from when I copy/pasted from webpack, I'll remove it.

index.html
funniray: I did double check, and I can confirm that the favicon does work as expect.

revam:
Themes are broken because it adds `/webui` to the css path.
image

revam:
Manually editing the url for the themes fixed it for the current tab, so the themes themselves are working as intended though.
image

harshithmohan:
```
'/webui/api/v3/WebUI/Theme.css': {
    target: 'http://localhost:8111',
    rewrite: path => path.replace('/webui', ''),
  },
  ```
  
@funniray   add this to proxy and it fixes the issue

funniray:
> add this to proxy and it fixes the issue
Was trying to avoid that solution, but I can't find anything better to be completely honest.

revam:
What if that is the non-hacky solution?

harshithmohan:
Well... including the Theme.css like that is the hack in the first place 😄

revam (approved):
:+1:

PR #563, opened 2023-06-10
fix: fix user management after server refactor

revam:
- fix the user avatar logic that changed in the last refactor on the server.

- added the missing AniDB user toggle. I don't know why it was missing.

- made sure the user is only logged out if they're revoking the API keys for themselves, and not for other users.

- made the `delete user` action clickable.

- made sure the edit and delete user buttons use the `pointer` cursor

- improved validation error toast messages

### DIFF
--- a/src/core/middlewares/rtkQueryError.ts
+++ b/src/core/middlewares/rtkQueryError.ts
@@ -1,8 +1,14 @@
 import type { Middleware, MiddlewareAPI } from '@reduxjs/toolkit';
 import { isRejectedWithValue } from '@reduxjs/toolkit';
+import { forEach } from 'lodash';
 
 import toast from '@/components/Toast';
 import Events from '../events';
+
+type ProblemDetails = { title: string; status: number; type: string; errors: Record<string, string[]>; traceId: string; };
+
+const isErrorObject = (target: any): target is ProblemDetails =>
+  typeof target === 'object' && target !== null && typeof target.title === 'string' && typeof target.errors === 'object' && target.errors !== null;
 
 const rtkQueryErrorLogger: Middleware = (_api: MiddlewareAPI) => dispatch => (action) => {
   if (isRejectedWithValue(action)) {
@@ -11,7 +17,20 @@ const rtkQueryErrorLogger: Middleware = (_api: MiddlewareAPI) => dispatch => (ac
       if (window.location.pathname !== '/webui/login') dispatch({ type: Events.STORE_CLEAR_STATE });
       toast.error('Unauthorized!', '', { toastId: 'unauthorized' });
     } else if (action.payload.originalStatus !== 503 && action.meta.arg.endpointName !== 'getRandomMetadata') {
-      toast.error(action.payload?.data?.title ?? action.payload.error ?? `${action.payload.status} - ${action.error.message}`, `Debug Info: ${action.meta.arg.endpointName}`);
+      const { data } = action.payload;
+      const endpointName = action.meta.arg.endpointName as string;
+      const method = action.meta.baseQueryMeta.request.method as string;
+      const url = new URL(action.meta.baseQueryMeta.request.url as string);
+      const endpoint = `${method} ${url.pathname + url.search}`;
+      if (isErrorObject(data)) {
+        if (data.title === 'One or more validation errors occurred.') {
+          forEach(Object.entries(data.errors), ([key, messageList]) =>
+            forEach(messageList, message =>
+              toast.error(message, `Error for key "${key}" during \`${endpoint}\` (${endpointName})`)));
+        }
+      } else {
+        toast.error(action.payload.error ?? `${action.payload.status} - ${action.error.message}`, `Error occured during \`${endpoint}\` (${endpointName})`);
+      }
     }
   }
 
--- a/src/core/rtkQuery/splitV3Api/userApi.ts
+++ b/src/core/rtkQuery/splitV3Api/userApi.ts
@@ -58,6 +58,15 @@ const userApi = splitV3Api.injectEndpoints({
       invalidatesTags: ['Users'],
     }),
 
+    // Remove the user.
+    deleteUser: build.mutation<void, number>({
+      query: userId => ({
+        url: `User/${userId}`,
+        method: 'DELETE',
+      }),
+      invalidatesTags: ['Users'],
+    }),
+
     // Change the password for the current user or a user by id.
     postChangePassword: build.mutation<void, { Password: string; RevokeAPIKeys: boolean; ID?: number; }>({
       query: ({ ID, ...body }) => ({
@@ -83,6 +92,7 @@ export const {
   useGetUsersQuery,
   usePostUserMutation,
   usePutUserMutation,
+  useDeleteUserMutation,
   usePostChangePasswordMutation,
   useGetCurrentUserQuery,
 } = userApi;
--- a/src/core/rtkQuery/splitV3Api/userApi.ts
+++ b/src/core/rtkQuery/splitV3Api/userApi.ts
@@ -16,7 +16,7 @@ const simplifyCommunitySites = (sites: Array<string>) => {
 
 const userApi = splitV3Api.injectEndpoints({
   endpoints: build => ({
-    // List all Users. Admin only
+    // List all users.
     getUsers: build.query<Array<UserType>, void>({
       query: () => ({ url: 'User' }),
       transformResponse: (response: Array<ApiUserType>) => {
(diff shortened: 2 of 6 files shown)


src/core/rtkQuery/splitV3Api/userApi.ts
```diff
@@ -30,54 +30,44 @@ const userApi = splitV3Api.injectEndpoints({
...
     }),
 
-    // Change the password for a user
-    postChangePassword: build.mutation<void, { Password: string, RevokeAPIKeys: boolean, userId: number, admin: boolean }>({
-      query: ({ admin, userId, ...body }) => ({
-        url: `User/${admin ? userId.toString() : 'Current'}/ChangePassword`,
-        body,
-        method: 'POST',
+    // Edit the current user or a user by id using a raw object to do a partial update.
+    putUser: build.mutation<UserType, { ID?: number | undefined } & PutUserType>({
```
revam, line 48: Using `ID` as the user id field to make it so the normal `UserType` can be used as the `PutUserType`.

src/core/types/api/user.ts
```diff
@@ -2,19 +2,28 @@ type BaseUserType = {
   ID: number;
   Username: string;
   IsAdmin: boolean;
-  RestrictedTags: Array<number>;
-  Avatar: {
-    RelativeFilepath: string;
-    ID: number;
-  };
+  RestrictedTags: number[];
```
revam, line 5: I can change back the `number[]` to `Array<number>` if you would rather have the full types for the primitive array values. @hidden4003

revam: Though personally I like to use the simplified array declarations when possible (meaning when the type in the array does not include any type parameters).

src/pages/settings/tabs/UserManagementSettings.tsx
```diff
@@ -133,15 +135,15 @@ function UserManagementSettings() {
 
   const changeAvatar = (avatar: File | undefined) => {
     if (!avatar) return;
-    editAvatar({ avatar, userId: selectedUser.ID })
-      .then(() => toast.success('', 'Avatar updated!'))
+    fileToDataUrl(avatar)
```
harshithmohan, line 138: If we don't have a changeAvatar endpoimt, we don't need these 2 functions right?
When we save the whole page it should be changed along with other user details

revam: As long as we can somehow convert the file to a data url and somehow reset it to an empty string, then no, we don't need these functions.

src/pages/settings/tabs/UserManagementSettings.tsx
```diff
@@ -67,15 +70,15 @@ function UserManagementSettings() {
   const tags = useGetAniDBTagsQuery({ pageSize: 0, excludeDescriptions: true });
 
   useEffect(() => {
-    if (users.length > 0) setSelectedUser(users[0]);
+    if (users.length > 0) setSelectedUser(cloneDeep(users[0]));
```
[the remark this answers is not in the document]
revam: Because otherwise we may be editing the data from the user list? Is it wrong to do a deep clone here?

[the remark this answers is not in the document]
revam: Should I revert it?

src/pages/settings/tabs/UserManagementSettings.tsx
revam: Also, if we plan to use a client-side cropping util of some sorts, then can this function be used to take the file and give it to the util (that hopefully opens up as a modal of some sorts).

src/pages/settings/tabs/UserManagementSettings.tsx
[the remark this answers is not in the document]
hidden4003: no actual need to clone, you only need to clone if you edit the array, but we also have immer so it is preferable to use that instead of cloning, as immer will take care of all the needed cloning automatically and prevent any errors in that regards, but here you don't need a deepClone.

src/core/types/api/user.ts
```diff
@@ -2,19 +2,28 @@ type BaseUserType = {
...
+type BasePostUserType = PutUserType & { Password?: string | null | undefined };
+
 export type CommunitySitesType = {
   AniDB: boolean;
   Trakt: boolean;
   Plex: boolean;
 };
 
-export type ApiUserType = BaseUserType & { CommunitySites: Array<string> };
+export type ApiUserType = BaseUserType & { CommunitySites: string[] };
```
hidden4003, line 19: Do we really need so many identical types just named differently? as when I read the code I get the impression that they are all different in some way, or is it to safeguard against calling the wrong hook type of thing? But then doesn't typescript just convert from one type to another if fields are the same? @harshithmohan is it warranted you think to have so many superficial types?

src/pages/settings/tabs/UserManagementSettings.tsx
```diff
@@ -133,15 +135,15 @@ function UserManagementSettings() {
...
       .catch(error => console.error(error));
   };
 
   const removeAvatar = () => {
-    deleteAvatar(selectedUser.ID)
-      .then(() => toast.success('', 'Avatar removed!'))
-      .catch(error => console.error(error));
+    // Setting the avatar to an empty string will tell the server to remove the
+    // avatar.
+    setSelectedUser(user => ({ ...user, Avatar: '' }));
```
hidden4003, line 148: and now here you do not use deep clone which would keep the community sites object at the same address which is fine I guess, but actually here deepClone would make more sense, then again please use immer for that and don't do it manually

hidden4003 (changes requested):

src/core/types/api/user.ts
harshithmohan: I don't think we need so many types....
If we make ID and Password optional in the `UserType`, it's more than enough, IMO

src/pages/settings/tabs/UserManagementSettings.tsx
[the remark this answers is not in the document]
harshithmohan: Data from `users` cannot really be edited but I get your point.
Also, we should probably change `selectedUser` to use Immer as avael said

src/core/middlewares/rtkQueryError.ts
```diff
@@ -1,17 +1,36 @@
 import type { Middleware, MiddlewareAPI } from '@reduxjs/toolkit';
 import { isRejectedWithValue } from '@reduxjs/toolkit';
+import { forEach } from 'lodash';
 
 import toast from '@/components/Toast';
 import Events from '../events';
 
+function isErrorObject(target: unknown): target is { title: string; traceId: string; type: string; status: 400; errors: Record<string, string[]>; } {
```
harshithmohan, line 8: Explain what this black magic is and why not
`isErrorObject(target: { title: string; traceId: string; type: string; status: 400; errors: Record<string, string[]> })`

src/core/rtkQuery/splitV3Api/userApi.ts
harshithmohan: Defeats the purpose of having a `PutUserType`
Just change ID to optional in `UserType` and use that

src/core/middlewares/rtkQueryError.ts
harshithmohan: Also, use `const function = () => {}` notation

src/pages/settings/tabs/UserManagementSettings.tsx
```diff
@@ -27,6 +27,15 @@ import {
 } from '@/core/rtkQuery/plexApi';
 import { useGetAniDBTagsQuery } from '@/core/rtkQuery/splitV3Api/tagsApi';
 
+function fileToDataUrl(file: File): Promise<string | never> {
```
harshithmohan, line 30:
```suggestion
const fileToDataUrl = (file: File): Promise<string | never> => {
```

harshithmohan: Also do we need to specify the return type?
It should be automatically inferred

src/pages/settings/tabs/UserManagementSettings.tsx
harshithmohan: nvm, I read the code wrong

harshithmohan (changes requested):

src/core/middlewares/rtkQueryError.ts
revam: It's a type guard. It tells typescript to assume `target` is the given type if the function returns true.

src/core/middlewares/rtkQueryError.ts
```diff
@@ -1,17 +1,36 @@
 import type { Middleware, MiddlewareAPI } from '@reduxjs/toolkit';
 import { isRejectedWithValue } from '@reduxjs/toolkit';
+import { forEach } from 'lodash';
 
 import toast from '@/components/Toast';
 import Events from '../events';
 
+type ProblemDetails = { title: string; status: number; type: string; errors: Record<string, string[]>; traceId: string; };
```
revam, line 8: It's named `ProblemDetails` since that's the name of the type on the server side.

src/core/rtkQuery/splitV3Api/userApi.ts
```diff
@@ -30,54 +30,53 @@ const userApi = splitV3Api.injectEndpoints({
...
 
-    // Change the password for a user
-    postChangePassword: build.mutation<void, { Password: string, RevokeAPIKeys: boolean, userId: number, admin: boolean }>({
-      query: ({ admin, userId, ...body }) => ({
-        url: `User/${admin ? userId.toString() : 'Current'}/ChangePassword`,
-        body,
-        method: 'POST',
+    // Edit the current user or a user by id using a raw object to do a partial update.
+    putUser: build.mutation<UserType, UserType>({
+      query: ({ ID, CommunitySites, Password: __, ...user }) => ({
```
revam, line 49: I made sure to omit the Password from the body to send so the server won't complain — even if we accidentally include it at some point — since our user type (`UserType`) allows it to included, but the server will reject the property since it's an unknown property from it's point of view. This is mostly because our type don't fully match what the server is expecting, but if you two are fine with it then I won't complain.

harshithmohan (approved):

hidden4003 (approved):

PR #585, opened 2023-07-09
Fix RemoveMissingFiles

Terrails:


revam (changes requested):
Can you also remove line 120 (`data: false`) and 127 (`data?: boolean;`), so `data` is gone for good?

Terrails:
> Can you also remove line 120 (`data: false`) and 127 (`data?: boolean;`), so `data` is gone for good?

Done

revam (approved):

PR #587, opened 2023-07-13
Fix `Unrecognized Files` Endpoint

GamingLiamStudios:


harshithmohan:
Those endpoints do exist on the latest daily server. The beta webui is not supposed to work stable server that is why it uses the new endpoints

PR #589, opened 2023-07-15
fix filename in ed2k hash

revam:


src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -296,7 +296,7 @@ function UnrecognizedTab() {
 
   const ed2kLinks = () => {
     const fileList = selectedRows.length > 0 ? selectedRows : files.List;
-    return fileList.map(file => `ed2k://|file|${file.Locations[0]?.RelativePath ?? '<missing file path>'}|${file.Size}|${file.Hashes.ED2K}|/`);
+    return fileList.map(file => `ed2k://|file|${file.Locations[0]?.RelativePath?.split(/[\\/]+/g).pop() ?? ''}|${file.Size}|${file.Hashes.ED2K}|/`);
```
hidden4003, line 299: You probably want to use `[\\/]{1}` here as otherwise it will capture multiple characters if it happens to have // for example as a first separator.

hidden4003 (changes requested):

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
revam: It doesn't matter for this use case, since we only want the last element in the resulting array (the file name). So the less elements there are the better, imo.

Should we still change it, or did i convince you it is fine as-is?

~~Also, if it's only one character then the `{1}` can be omitted~~.

hidden4003 (approved):

PR #600, opened 2023-07-22
refactor: move to v3 queue api

revam:


revam:
The "Images" queue is internally named just "Image", and the v3 queue API just uses the queue names as-is, so had to adapt that on the client side. It now uses the key instead of the name for the operation.

revam:
Also there is a "Clear All" endpoint now if we want to add a button for that @ElementalCrisis

hidden4003 (approved):

PR #607, opened 2023-07-23
Fix Manually Linking Util

revam:


revam:
The logic to save is still a bit buggy in regards to RTK's cache. Will look at it some more tomorrow maybe, unless any of you know how to fix this.

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesTab.tsx
```diff
@@ -232,33 +242,108 @@ function LinkFilesTab() {
...
-      forEach(fileIds, (link) => {
-        if (link.FileID === 0) {
-          return;
+  useEffect(() => {
+    if (!isLinking || isLinkingRunning || getAnidbSeriesQuery.isLoading || !(getAnidbSeriesQuery.isSuccess && getAnidbSeriesQuery.data.ShokoID)) {
+      return;
+    }
+
+    const seriesID = getAnidbSeriesQuery.data.ShokoID;
+    const linkFiles = async () => {
```
hidden4003, line 251: seems to me that linkFiles does not have to be inside the useEffect it only makes it look a lot more complicated than it should be

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesTab.tsx
```diff
@@ -232,33 +242,108 @@ function LinkFilesTab() {
...
+      if (isLinkingRunning) return;
```
hidden4003, line 252: This is duplicate of top condition

hidden4003: Some explanation on what feature is being added here would be nice.

revam: > Some explanation on what feature is being added here would be nice.

We're making sure the Shoko series is only created when the 'Save' button is pressed, and not before that.

revam:
I was thinking of adding some "fancy" series guessing (spoiler; ~~it's regular expressions~~) while I was at it, and maybe some "fancy" episode number guessing too (~~using the same regexes~~).

hidden4003:
It looks a lot simpler

revam:
I'll do the series guessing and episode number guessing in another PR.

hidden4003 (approved):

PR #610, opened 2023-07-29
Fix Dashboard resizing issue on initial load.

ElementalCrisis:
Fixes the issue with the dashboard not being resized properly on initial load.

### DIFF
--- a/src/core/router/index.tsx
+++ b/src/core/router/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 
 import { createBrowserRouter, createRoutesFromElements, RouterProvider } from 'react-router-dom';
 import { Navigate, Route } from 'react-router';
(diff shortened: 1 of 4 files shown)


src/core/router/index.tsx
```diff
@@ -123,16 +123,22 @@ const Router = () => {
 
   const settingsQuery = useGetSettingsQuery(undefined, { skip: apikey === '' });
   const { theme } = settingsQuery.data?.WebUI_Settings ?? initialSettings.WebUI_Settings;
+  const bodyRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
     document.body.className = `${apikey === '' ? globalThis.localStorage.getItem('theme') : (webuiPreviewTheme ?? theme)} theme-shoko-gray`;
-    setTimeout(() => {
-      document.getElementsByTagName('body')[0].style.display = 'initial';
+    const timeoutId = setTimeout(() => {
```
hidden4003, line 130: Why the magic number and timeout? Is it possible to just have something like this?
```
useEffect(() => {
 if (bodyRef.current) {
        bodyRef.current.style.visibility = 'initial';
      }
}, []);
```

ElementalCrisis: We need the timeout to stop the flashing on the login page and 125 is the lowest I was able to go to achieve this while still being snappy. 

https://github.com/ShokoAnime/Shoko-WebUI/assets/9443295/d7a32fe1-d55f-4699-a1f4-52f30e8926ea

PR #611, opened 2023-07-29
fix: fix image placeholders

revam:
- change the background image placeholder component to accept image metadata instead of an url.

- add three separeate placeholders depening the state of the provided image metadata.

- refactored some parts to fix react breaking because of the use of hooks inside non-components (that are now made into "proper" components).


revam:
I'm ready to be roasted by @hidden4003  and @harshithmohan.

revam:
for context;

This PR fixes the placeholder text…

- …when there is no metadata available, e.g.;
image

- …when the image is not available in the image cache, but metadata exists, .e.g.;
image

- …when the image failed to load, but exists on the server, e.g.;
image

harshithmohan (approved):

hidden4003 (approved):

PR #616, opened 2023-08-07
Add Indicator For New News Post.

ElementalCrisis:
A simple indicator if there is a new news post.

image

### DIFF
--- a/src/pages/dashboard/panels/ShokoNews.tsx
+++ b/src/pages/dashboard/panels/ShokoNews.tsx
@@ -9,9 +9,21 @@ import { DashboardNewsType } from '@/core/types/api/dashboard';
 
 import { useGetShokoNewsFeedQuery } from '@/core/rtkQuery/externalApi';
 
+function isWithinXDaysFromToday(date: string) {
+  const itemDate = new Date(date);
+  const currentDate = new Date();
+  const oneDayInMilliseconds = 24 * 60 * 60 * 1000;
+  const differenceInMilliseconds = currentDate.getTime() - itemDate.getTime();
+  const differenceInDays = Math.floor(differenceInMilliseconds / oneDayInMilliseconds);
+  return differenceInDays <= 14;
+}
+
 const NewsRow = ({ item }: { item: DashboardNewsType }) => (
   <div className="flex flex-col" key={item.title}>
-    <p className="font-semibold">{item.date_published}</p>
+    <div className="flex gap-x-4">
+      <p className="font-semibold">{item.date_published}</p>
+      <p className="font-semibold text-panel-important">{isWithinXDaysFromToday(item.date_published) && 'New Post!'}</p>
+    </div>
     <a href={item.link} rel="noopener noreferrer" target="_blank" className="flex text-panel-primary font-semibold items-center mt-1 space-x-2 hover:text-panel-primary-hover">
       <p className="font-semibold">{item.title}</p>
       <Icon path={mdiOpenInNew} size={1} />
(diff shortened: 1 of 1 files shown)


src/pages/dashboard/panels/ShokoNews.tsx
```diff
@@ -9,9 +9,21 @@ import { DashboardNewsType } from '@/core/types/api/dashboard';
 
 import { useGetShokoNewsFeedQuery } from '@/core/rtkQuery/externalApi';
 
+function isWithinXDaysFromToday(date: string) {
```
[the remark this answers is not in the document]
harshithmohan: Also either change the function name to `isWithin14DaysFromToday` or add the number of days as an argument with 14 as the default parameter.
It doesn't matter much because we are using it in a single place, but just to be correct.

harshithmohan (changes requested):

src/pages/dashboard/panels/ShokoNews.tsx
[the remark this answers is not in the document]
ElementalCrisis: Originally, I was going to make it so you could specify how many days but then decided to hardcode it.

harshithmohan (approved):

src/pages/dashboard/panels/ShokoNews.tsx
```diff
@@ -9,9 +9,21 @@ import { DashboardNewsType } from '@/core/types/api/dashboard';
...
+  const oneDayInMilliseconds = 24 * 60 * 60 * 1000;
+  const differenceInMilliseconds = currentDate.getTime() - itemDate.getTime();
+  const differenceInDays = Math.floor(differenceInMilliseconds / oneDayInMilliseconds);
+  return differenceInDays <= 14;
+};
+
 const NewsRow = ({ item }: { item: DashboardNewsType }) => (
   <div className="flex flex-col" key={item.title}>
-    <p className="font-semibold">{item.date_published}</p>
+    <div className="flex gap-x-4">
```
harshithmohan, line 23: Just a little thing for the future. This is still fine, you can leave it.
As both the child elements have the class `font-semibold`, you could have just added the class here for the div and removed them from the children

src/pages/dashboard/panels/ShokoNews.tsx
```diff
@@ -9,9 +9,21 @@ import { DashboardNewsType } from '@/core/types/api/dashboard';
...
+      <p className="font-semibold">{item.date_published}</p>
+      <p className="font-semibold text-panel-important">{newNewsCheck(item.date_published) && 'New Post!'}</p>
```
harshithmohan, line 25: This works but the p element is still added to the DOM. Normally you would write
`{newNewsCheck(item.date_published) && (<p className="font-semibold text-panel-important">'New Post!'</p>)}`

hidden4003: Yea the checks needs to include the whole element not just the text

src/pages/dashboard/panels/ShokoNews.tsx
```diff
@@ -9,9 +9,21 @@ import { DashboardNewsType } from '@/core/types/api/dashboard';
 
 import { useGetShokoNewsFeedQuery } from '@/core/rtkQuery/externalApi';
 
+const newNewsCheck = (date: string) => {
+  const itemDate = new Date(date);
```
hidden4003, line 13: Please use moment library for this.

src/pages/dashboard/panels/ShokoNews.tsx
hidden4003: Arguably, just make a function component inline (in same file) and return either null or p depending on the condition, it will not have the curly braces in JSX which I don't like. For reference like OptionButton here https://github.com/ShokoAnime/Shoko-WebUI/blob/edc1645fb196064ac2d5039adea1bf58df0af123/src/pages/collection/Collection.tsx#L26-L30
It would take a newsitem or just the date and then either return a `<p>` or a null.

hidden4003 (changes requested):

hidden4003 (approved):

PR #619, opened 2023-08-10
Add dprint config, update eslint config

harshithmohan:
~~Thinking of adding pre-commit hooks, and need to check if dprint checks on GH are working properly.
That's why this is in a draft state~~


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
hidden4003, line 1: does it work in windows ?

harshithmohan: It does on mine... Not sure how

PR #620, opened 2023-08-11
Add queue modal

revam:


revam:
The ` disabled` queue item property sent from the current server is broken and will be fixed by https://github.com/ShokoAnime/ShokoServer/pull/1079, and the api endpoint can be optimised so it doesn't freeze the whole server (because of a DB lock) while getting the queue items, but that's for later. The client-side modal is working as expected besides those two points.

src/components/Dialogs/QueueModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  const handleToggleQueue = useEventCallback(async () => {
+    await queueOperation({ operation: isPaused ? 'Start' : 'Stop', queue: activeTab });
+  });
+
+  // We're intentionally not letting RTK invalidate any tags for automagic query
+  // updates, since if we did then it would fetch in the background when the
+  // modal is not in view, since it's technically always mounted, just not
+  // shown.
+  useLayoutEffect(() => {
```
harshithmohan, line 169: why `useLayoutEffect` instead of `useEffect`?

harshithmohan (changes requested):

src/components/Dialogs/QueueModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    const itemArray = map(array, item => (
+      <div className="mt-2 flex gap-x-3" key={`item-${item.ID}`}>
+        <div className="flex-grow">
+          {item.Name}
+        </div>
+        <div className={cx(['px-4', item.IsRunning ? 'text-panel-important' : undefined, item.IsDisabled ? 'text-panel-warning' : undefined])}>
+          {/* eslint-disable-next-line no-nested-ternary */}
+          <Icon path={item.IsRunning ? mdiRun : item.IsDisabled ? mdiAlertCircleOutline : mdiHelpCircleOutline} size={1} />
```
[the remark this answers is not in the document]
revam: I'll update it, but to me does the first one look cleaner.

src/components/Dialogs/QueueModal.tsx
```diff
@@ -0,0 +1,234 @@
+import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { useSelector } from 'react-redux';
+import cx from 'classnames';
+import { filter, map, reduce, throttle } from 'lodash';
+import { Icon } from '@mdi/react';
+import { mdiAlertCircleOutline, mdiCheckboxMarkedCircleOutline, mdiChevronRight, mdiCircleOutline, mdiHelpCircleOutline, mdiLoading, mdiRun } from '@mdi/js';
+import { useEventCallback } from 'usehooks-ts';
```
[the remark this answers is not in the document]
revam: for any lurkers; https://blog.logrocket.com/what-you-need-know-react-useevent-hook-rfc/

src/core/slices/mainpage.ts
```diff
@@ -8,6 +8,7 @@ type State = {
   };
   queueStatus: QueueStatusType;
   banStatus: AniDBBanType;
+  queueModalOpen: boolean;
```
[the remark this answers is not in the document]
revam: Because then we would have two modals mounted to the react tree on the dashboard, and one modal everywhere else. The modal is only mounted in the top-nav component, but the dispatch to open it is in two components, the top-nav and the dashboard queue processor widget.

Also the icon in the top-nav will light up when you open the modal from the dashboard widget if we store the state in a shared place. We won't be able to do that if we only store the state locally in the parent component of the modal.

src/components/Dialogs/QueueModal.tsx
revam: Because otherwise it will render before it should. E.g. it shows the empty queue page, then the loader, then the items, instead of just showing the loader then the items, also the same issues while switching tabs. If you have a better way that also works with `useEffect`  instead of `useLayoutEffect` then I'm all ears.

src/core/slices/mainpage.ts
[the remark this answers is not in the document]
harshithmohan: Fair

src/components/Dialogs/QueueModal.tsx
harshithmohan: I'm confused... useLayoutEffect runs *before* useEffect

revam: Also, we're talking in terms of milliseconds, or a flicker on the screen, when i say it's showing the empty queue page (when it shouldn't).

src/components/Dialogs/QueueModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+import ModalPanel from '@/components/Panels/ModalPanel';
+import { useGetQueueOperationMutation, useLazyGetQueueItemsQuery } from '@/core/rtkQuery/splitV3Api/queueApi';
+import { RootState } from '@/core/store';
+import { QueueItemType } from '@/core/types/api/queue';
+import Input from '../Input/Input';
+import MenuButton from '../Utilities/Unrecognized/MenuButton';
+
+type QueueName = keyof typeof names;
```
harshithmohan, line 16: Usage of variable before it's declared.
How did eslint not complain? We'll never know
fix it

harshithmohan (changes requested):

harshithmohan (approved):

src/components/Dialogs/QueueModal.tsx
hidden4003: useLayoutEffect is for libraries as far as I remember, also docs say 

> The code inside useLayoutEffect and all state updates scheduled from it block the browser from repainting the screen. When used excessively, this makes your app slow. When possible, prefer [useEffect.](https://react.dev/reference/react/useEffect)

hidden4003 (approved):

PR #624, opened 2023-08-13
Misc changes

revam:


revam:
Ready for review. A bunch of small and large changes all over the place. Everything is (humanly) tested.

src/components/Collection/AnidbDescription.tsx
```diff
@@ -1,38 +1,38 @@
...
-);
+const CleanMiscLinesRegex = /^(\*|--|~) /sg;
 
-const AnidbDescription = ({ text }) => {
-  const lines = [] as Array<JSX.Element>;
+const CleanMultiEmptyLinesRegex = /\n{2,}/sg;
+
+const LinkRegex = /(?<url>http:\/\/anidb\.net\/(?<type>ch|cr|[feat])(?<id>\d+)) \[(?<text>[^\]]+)]/g;
+
+const AnidbDescription = ({ text }: { text: string | null | undefined }) => {
```
harshithmohan, line 11: You don't need to specify `null | undefined`

src/components/Dialogs/DeleteFilesModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { mdiMinusCircleOutline } from '@mdi/js';
+import Icon from '@mdi/react';
+import { useEventCallback } from 'usehooks-ts';
+
+import Button from '../Input/Button';
+import ModalPanel from '../Panels/ModalPanel';
+
+import type { FileType } from '@/core/types/api/file';
+
+interface Props {
```
harshithmohan, line 11: use `Type` not `Interface`

src/core/middlewares/signalr.ts
```diff
@@ -11,9 +11,12 @@ import moment from 'moment';
 
 import Events from '@/core/events';
 import { splitV3Api } from '@/core/rtkQuery/splitV3Api';
-import { setFetched, setHttpBanStatus, setQueueStatus, setUdpBanStatus } from '@/core/slices/mainpage';
+import { setFetched, setHttpBanStatus, setNetworkStatus, setQueueStatus, setUdpBanStatus } from '@/core/slices/mainpage';
+import { AVDumpEventTypeEnum } from '@/core/types/signalr';
 
-import type { AniDBBanItemType } from '@/core/types/signalr';
+import { restoreAVDumpSessions, updateAVDumpEvent } from '../slices/utilities/avdump';
```
harshithmohan, line 17:
```suggestion
import { restoreAVDumpSessions, updateAVDumpEvent } from '@/core/slices/utilities/avdump';
```

harshithmohan: Change and run `eslint --fix`

src/core/rtkQuery/splitV3Api/fileApi.ts
```diff
@@ -31,11 +25,12 @@ const fileApi = splitV3Api.injectEndpoints({
...
 
     // Run a file through AVDump and return the result.
-    postFileAVDump: build.query<AVDumpResultType, number>({
+    postFileAVDump: build.mutation<void, number>({
       query: fileId => ({
-        url: `File/${fileId}/AVDump`,
+        url: `File/${fileId}/AVDump?immediate=false`,
         method: 'POST',
       }),
+      invalidatesTags: ['AVDumpEvent'],
```
harshithmohan, line 33: If you have set `immediate=false`, how is invalidation supposed to work?
`getFiles` query will run before dumping is complete

src/core/slices/utilities/avdump.ts
```diff
@@ -1,27 +1,87 @@
-import { createSlice } from '@reduxjs/toolkit';
-import { omit } from 'lodash';
+import { type PayloadAction, createSlice } from '@reduxjs/toolkit';
```
harshithmohan, line 1: Keep type import separate as before for uniformity

src/core/types/api/file.ts
```diff
@@ -55,6 +57,21 @@ export type FileAniDBType = {
...
+  Status: 'Running' | null;
+  Progress: number;
+  SucceededCreqCount: number;
+  FailedCreqCount: number;
+  PendingCreqCount: number;
+  StartedAt: string;
+  LastDumpedAt: string | null;
+  LastVersion: string | null;
+};
+
```
harshithmohan, line 74: Why two types?
Combine them and keep the extra ones optional.
Because this will cause issues

src/core/types/signalr.ts
```diff
@@ -33,3 +33,128 @@ export type AniDBBanType = {
...
+  videoIDs: number[];
+  commandID: number | null;
+  progress: number;
+  succeededCreqCount: number;
+  failedCreqCount: number;
+  pendingCreqCount: number;
+  message: string;
+  startedAt: string;
+  endedAt: string;
+};
```
harshithmohan, line 141: How sure are you that this will work without issues?

src/pages/collection/series/SeriesFileSummary.tsx
```diff
@@ -106,29 +185,51 @@ const SeriesFileSummary = () => {
...
+          </option>
+          <option value="GroupName,FileVersion,FileSource,FileLocation,VideoCodecs,VideoBitDepth,VideoResolutuion,AudioCodecs,AudioLanguages,AudioStreamCount">
+            + Audio
+          </option>
+          <option value="GroupName,FileVersion,FileSource,FileLocation,VideoCodecs,VideoBitDepth,VideoResolutuion,AudioCodecs,AudioLanguages,AudioStreamCount,SubtitleCodecs,SubtitleLanguages,SubtitleStreamCount">
+            + Subtitles
+          </option>
+        </Select>
+        {map(summary.SourceByType, ({ count, source, type }, index) => (
+          <Fragment key={`${type}-${index}`}>
```
harshithmohan, line 209: use `React.Fragment` for readability

src/pages/utilities/UnrecognizedUtility.tsx
```diff
@@ -67,6 +68,14 @@ function UnrecognizedUtility() {
...
         },
       }),
+      columnHelper.display({
+        id: 'status',
+        header: 'Status',
+        cell: info => <AVDumpFileIcon file={info.row.original} />,
+        meta: {
+          className: 'w-20',
+        },
+      }),
```
harshithmohan, line 78: This is supposed to added in the UnrecognizedTab, not here

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -1,29 +1,25 @@
...
   mdiMinusCircleOutline,
   mdiOpenInNew,
   mdiRefresh,
 } from '@mdi/js';
 import { Icon } from '@mdi/react';
-import { createColumnHelper, getCoreRowModel, getFilteredRowModel, useReactTable } from '@tanstack/react-table';
-import cx from 'classnames';
-import { forEach } from 'lodash';
-import { useDebounce } from 'usehooks-ts';
+import { type Table, getCoreRowModel, getFilteredRowModel, useReactTable } from '@tanstack/react-table';
```
harshithmohan, line 18:
```suggestion
import type { Table, getCoreRowModel, getFilteredRowModel, useReactTable } from '@tanstack/react-table';
```

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -76,7 +70,27 @@ const Menu = (
...
+  const removeFileFromSelection = useEventCallback((fileId: number) => {
+    // Figure this out.
+    some(tableSelectedRows.rows, (row) => {
+      if (row.original.ID === fileId) {
+        row.toggleSelected(false);
+        return true;
+      }
+      return false;
+    });
+  });
```
harshithmohan, line 90: I'm all for using new library functions but a `forEach` and `break` is more easily understandable by majority of the people

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -115,9 +129,9 @@ const Menu = (
 
     if (failedFiles) toast.error(`Rehash failed for ${failedFiles} files!`);
     if (failedFiles !== fileList.length) toast.success(`Rehashing ${fileList.length} files!`);
-  };
+  });
 
-  const rescanFiles = (selected = false) => {
+  const rescanFiles = useEventCallback((selected = false) => {
```
harshithmohan, line 134: Spamming `useEventCallback` makes the app slower, not faster.
Are you sure about all of this?

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -169,15 +183,21 @@ const Menu = (
...
     </div>
   );
 };
 
 function UnrecognizedTab() {
   const navigate = useNavigate();
 
-  const { columns: tempColumns } = useUnrecognizedUtilityContext();
-
+  const { columns } = useUnrecognizedUtilityContext();
```
harshithmohan, line 200: again, "Status" column was supposed to be added here

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -299,51 +224,36 @@ function UnrecognizedTab() {
...
-  const avdumpFiles = async () => {
-    for (let i = 0; i < selectedRows.length; i += 1) {
-      // Files cannot be dumped in parallel on the server, yet
-      // Use Promise.all when avdump3 is implemented and parallel dumping is possible
-      if (!avdumpList[selectedRows[i].ID]?.hash) {
-        // eslint-disable-next-line no-await-in-loop
-        await runAvdump(selectedRows[i].ID);
+  const selectedRows = useMemo(() => tableSelectedRows.rows.map(row => row.original), [tableSelectedRows]);
+  const isAvdumpFinished = useMemo(() => every(selectedRows, row => row.AVDump.LastDumpedAt), [selectedRows]);
+  const dumpInProgress = some(avdumpList.sessions, session => session.status === 'Running');
```
harshithmohan, line 229: Note: This is a good place to use the `some` function. The other place is not it

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -299,51 +224,36 @@ function UnrecognizedTab() {
...
+    if (isAvdumpFinished && !dumpInProgress) {
+      setSeriesSelectModal(true);
+    } else {
+      try {
+        await Promise.all(map(selectedRows, async (row) => {
+          if (row && !row.AVDump.LastDumpedAt && !row.AVDump.Status) {
+            // eslint-disable-next-line no-await-in-loop
+            await fileAvdumpTrigger(row.ID);
+          }
+        }));
```
harshithmohan, line 241: ... does this even work?
You are awaiting an await.. That's not how `Promise.all` works

src/core/types/signalr.ts
harshithmohan: image

Union operator doesn't work like you think it does

harshithmohan (changes requested):

src/components/Dialogs/DeleteFilesModal.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useEffect, useMemo } from 'react';
+import { mdiMinusCircleOutline } from '@mdi/js';
+import Icon from '@mdi/react';
+import { useEventCallback } from 'usehooks-ts';
+
+import Button from '../Input/Button';
+import ModalPanel from '../Panels/ModalPanel';
```
harshithmohan, line 7:
```suggestion
import Button from '@/components/Input/Button';
import ModalPanel from '@/components/Panels/ModalPanel';
```

harshithmohan (changes requested):

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
hidden4003: In this case, some is a short circuited forEach but if you return false from normal forEach it is same behavior, I usually use forEach with return false.

hidden4003: Also it only deletes 1 file it seems, is that the intent?

src/core/rtkQuery/splitV3Api/fileApi.ts
revam: Immediate false is to make it queue the item instead of running it then and there. It defulaults to true as to not break web ui v1. The tag invalidation is to get theblist to update with the new information that the command is queued (or running) for the file(s).

src/core/types/api/file.ts
revam: So we know which fields are populated when.

src/core/types/signalr.ts
revam: Change `seriesId` to `seriesID` and add `as const` at the end of the declaration of `testObj` to get it to qualify as a progress event.

Also, these types are sent from the server, and knowing which fields are available for which events are useful, since we can use the `type` field to narrow down the availablefields using typescript type guards. If i wanted to change it, then i would declare each event separately and export them, then create a union of the now named events.

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
revam: Look closelier.

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
revam: It removes the file from the list, it doesn't delete it. Also didn't know about the fact that we can `return false` in a `forEach` loop (in lodash) to get it to terminate early.

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
revam: I have no idea what you mean. It works just fine. It could be changed to just return the promise from the trigger instead of awaiting it, if that's what you mean.

src/components/Collection/AnidbDescription.tsx
revam: It is `text: string | null | undefined` because the value passed to it is maybe a string, maybe `null` or maybe `undefined`. It is not `text?: string | null | undefined` because `text` in mandatory.

Context:
image
image

src/core/types/api/file.ts
revam: Examples;
image
image

revam: Fudge. I spot an error.

src/core/types/signalr.ts
revam: Your example;
image
image

W/ what i just suggested;
image
image

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
revam: Simplified.

src/components/Collection/AnidbDescription.tsx
harshithmohan: Why pass null or undefined values intentionally? You could have a fallback for the second screenshot -> `?? ''`

revam: I forgot about that.

src/core/rtkQuery/splitV3Api/fileApi.ts
harshithmohan: The files api gives the information that the file is queued?

src/components/Collection/AnidbDescription.tsx
revam: Fixed.

harshithmohan (approved):

src/core/rtkQuery/splitV3Api/fileApi.ts
revam: every file has an avdump section sent with it now, and in it we get info about if the file has been successfully dumped before, if a command is session is queued or if a session is running. also if a session is running then additional data is available from the session.

PR #636, opened 2023-08-20
feat: add auto-search to avdump copy modal

revam:


src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -27,23 +39,18 @@ function AvDumpSeriesSelectModal({ getLinks, onClose, show }: Props) {
     }, 200),
   ).current;
 
-  const handleSearch = (query: string) => {
+  const handleSearch = useEventCallback((query: string) => {
```
revam, line 42: Before anyone else comments; I added this to keep the effect from re-running on every render since the identity of the function was _unstable_ before this. With this change will the new effect I added below only update the search when the common series changes, which in turn will only change when the links change, which in turn will only change when the modal is opened (or closed).

harshithmohan: It would be helpful if you could should what happens when it is unstable vs when it's not.

hidden4003 (approved):

PR #639, opened 2023-08-27
misc: another update to the regexes

revam:
another update to the regexes to account for EDs/OPs using decimal point numbering


revam:
Forgot to say in the commit message, but there are no known regressions for adding this.

hidden4003 (approved):

PR #646, opened 2023-09-21
fix: fix auto match regexes crashing safari <16.4

revam:
fix auto match regexes crashing safari <16.4 by encapsulating the logic to create the array holding the regexes in a try…catch block.

### DIFF
--- a/src/core/utilities/auto-match-regexes.ts
+++ b/src/core/utilities/auto-match-regexes.ts
@@ -1,95 +1,101 @@
 import type { PathDetails, PathMatchRule } from './auto-match-logic';
 
-// TODO: mohan couldn't find a dprint setting to make the = go to next line. This needs to be fixed.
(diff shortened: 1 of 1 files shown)


src/core/utilities/auto-match-regexes.ts
```diff
@@ -1,95 +1,101 @@
...
+      transform: defaultTransform,
+    },
+    // TODO: Add more rules here.
+  );
+} catch (error) {
+  console.error(error);
```
hidden4003, line 98: How does the fix work? Auto match won't work on those safari browsers that crash on the regexes?

revam: yes

revam: also eslint was complaining about the empty catch block, so i added a console.error statement just because.

PR #647, opened 2023-09-22
First run update

harshithmohan:


src/core/patches.ts
```diff
@@ -7,4 +7,5 @@ export const webuiSettingsPatches = {
     else webuiSettings.updateChannel = 'Stable';
     return { ...webuiSettings, settingsRevision: 5 };
   },
-} as Record<number, (any) => WebUISettingsType>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+} as Record<number, (oldWebuiSettings: any) => WebUISettingsType>;
```
hidden4003, line 11: why not replace with never?

hidden4003 (approved):

src/core/patches.ts
harshithmohan: because...
image

hidden4003: wait, why is it any if it is actually WebUISettingsType?

harshithmohan: `WebUISettingsType` is updated everytime settings change.
In this case, if we make it `WebUISettingsType`, it would say `updateChannel === 'unstable'` will always be false since it cannot be 'unstable' in the new type

harshithmohan: and now I'm confused why there is a `as WebUISettingsType`.....

harshithmohan: image
I have no idea now... 😭

hidden4003: > `WebUISettingsType` is updated everytime settings change.
You mean it is a dynamic type? But I see that it is not. https://github.com/ShokoAnime/Shoko-WebUI/blob/a53887a76ddae1a5f12cfebefaa6c447adc7ef09/src/core/types/api/settings.ts#L177-L186
To me it looks like you meant to compare with a version of settings that Webui v1 used, so maybe in needs a new type?
Thing that confuses me is this line:
`const webuiSettings = oldWebuiSettings as WebUISettingsType;`
If you type cast it to WebUISettingsType anyways why can't you declare that oldWebuiSettings is that from the start not keep it as any type.

hidden4003: Maybe you need a separete type for legacy webui v1 settings and put a function like `isErrorObject` to typecast it to that type, so you know that at that point you have a legacy settings object not v2 one. https://github.com/ShokoAnime/Shoko-WebUI/blob/a53887a76ddae1a5f12cfebefaa6c447adc7ef09/src/core/middlewares/rtkQueryError.ts#L17-L19

harshithmohan: We can remove the `as WebUISettingsType`. It has to be any and not a legacy webui v1 type.
The patches thing exists for migration, that migration can be even between v2 versions, like from v2.1.0 to v2.1.1.
We can't keep adding type for every settings model change

hidden4003: Sure that works too.

harshithmohan: Done

PR #649, opened 2023-10-19
Update collection to 7.5

harshithmohan:


src/pages/collection/Collection.tsx
```diff
@@ -18,18 +29,79 @@ const OptionButton = ({ icon, onClick }) => (
...
+          </div>
+          <div className="line-clamp-2">{series.Name}</div>
+        </div>
+      </div>
+    </Link>
+  );
+};
+
+const TimelineSidebar = ({ series }: { series: SeriesType[] }) => (
+  <div className="flex min-h-full w-[26.125rem] overflow-hidden transition-all">
```
ElementalCrisis, line 59: Shouldn't the width be 28.125rem as this width is including the gap between the two panels?

harshithmohan: The next line has `ml-8` which is the gap

harshithmohan: @ElementalCrisis Fixed

src/components/Collection/CollectionView.tsx
```diff
@@ -8,56 +8,102 @@ import { useVirtualizer } from '@tanstack/react-virtual';
...
 type Props = {
   mode: string;
   setGroupTotal: (total: number) => void;
+  setTimelineSeries: (series: SeriesType[]) => void;
+  type: 'collection' | 'group';
+  isSidebarOpen: boolean;
 };
 
-const pageSize = 50;
+const defaultPageSize = 50;
```
hidden4003, line 28: Should we move this to settings as well, seeing that you added settings query in here anyways?

src/core/rtkQuery/splitV3Api/collectionApi.ts
```diff
@@ -47,8 +50,21 @@ const collectionApi = splitV3Api.injectEndpoints({
         params: { includeEmpty, topLevelOnly },
       }),
     }),
-    getGroupSeries: build.query<SeriesType[], { groupId?: string }>({
-      query: ({ groupId }) => ({ url: `Group/${groupId}/Series` }),
+    getGroupSeries: build.query<InfiniteResultType<SeriesType[]>, { groupId?: string, randomImages?: boolean }>({
```
hidden4003, line 53: Should probably be `groupId: string`, not `?: string`

hidden4003 (approved):

src/components/Collection/CollectionView.tsx
harshithmohan: The GroupView endpoint only allows 100 group IDs so it would have to be a setting with a max limit of 100.. and people will complain because 90% of the people who would change the setting would want to be infinite size. We can add that setting afterwards.

src/core/rtkQuery/splitV3Api/collectionApi.ts
harshithmohan: done

src/components/Collection/CollectionView.tsx
hidden4003: Well on the other hand we have infinite load anyway so page size doesn't really matter.

PR #654, opened 2023-10-22
QoL Improvements 

ElementalCrisis:


src/pages/collection/Series.tsx
```diff
@@ -106,13 +106,17 @@ const Series = () => {
...
-                  to={`/webui/collection/group/${series.IDs?.ParentGroup}`}
-                >
-                  {group.Name}
-                </Link>
-                <Icon path={mdiChevronRight} size={1} />
+                {group.Size > 1 && (
+                  <>
+                    <Link
+                      className="font-semibold text-panel-text-primary"
+                      to={`/webui/collection/group/${series.IDs?.ParentGroup}`}
```
hidden4003, line 113: I think `?.ParentGroup` check needs to be moved into `group.Size > 1` condition otherwise you will be rendering a broken link which doesn't make a lot of sense.

hidden4003 (approved):

src/pages/collection/Series.tsx
harshithmohan: `group` will be empty if `ParentGroup` doesn't exist because the query will be skipped

hidden4003: I mean if you anticipate that `${series.IDs?.ParentGroup}` might be null then just move it to the line 109 and don't try to render a link at all if you are unable to make a valid route anyways.

ElementalCrisis: I don't think the optional chaining there is needed as every series will always have a parent group. Its why I added the ``group.Size`` check in the first place since without it, you could go into the group for single series. 

image

harshithmohan: There, I fixed it 👍 
`IDs` will always exist so there was no point in putting the `?` anyway

PR #659, opened 2023-10-25
Makes Series Background Image Actually Random.

ElementalCrisis:
seriesData only returns the default fanart for a series. Used the same endpoint we use in Settings to get series fanart and truly return a random one.


harshithmohan (approved):

hidden4003:
I disagree, series should only show random fanart from same series not just any fanart.

harshithmohan:
Eh.. It does do that
`const imagesData = useGetSeriesImagesQuery({ seriesId: seriesId! }, { skip: !seriesId });`
Gets random images from the same series

PR #666, opened 2023-10-30
fix: fix anidb description… again

revam:
- add missing link templates to regex, since we only accounted for short links, but people also insert long links,

- merge multi spaces into single space,

- remove summary credits,

- account for people's disability to spell correctly, and

- fix tag descriptions.


revam:
Just noticed, but this PR got the ~~un~~lucky id of 666.

PR #672, opened 2023-11-10
Recently Imported Episodes Panel: Guarantee a Unique Key Per Element

Queuecumbr:
This fixes an edge case where duplicate keys may end up used in instances where a regular episode and an O episode is imported. Take my case for example:

image

Both O1 and O2 have the file IDs 10126 and 10127 respectively. However E1 would have both 10126 and 10127 linked to it. So when rendering O1, O2, and both instances of E1, duplicate keys will be used since it was based on the FileID.

The following warning will be displayed when running the code in development mode as a result:
Screenshot 2023-11-10 at 3 12 51 PM

This pull request should fix this issue by setting the key to EpisodeID-FileID instead of just to FileID.


revam (approved):
LGTM 👍

ElementalCrisis:
Could have sworn I fixed this, thanks for the PR!

PR #674, opened 2023-11-11
Update Series Overview to New Design.

ElementalCrisis:
Updates Series Overview to match new design.


src/components/Collection/SeriesInfo.tsx
```diff
@@ -0,0 +1,187 @@
...
+            )
+            : '--'}
+        </div>
+        <div className="flex justify-between capitalize">
+          <div className="font-semibold">Studio</div>
+          <div>{overview?.Studios?.[0] ? overview?.Studios?.[0].Name : 'Studio Not Listed'}</div>
+        </div>
+      </div>
+
+      {/*   <div className="flex flex-col gap-y-4"> */}
```
hidden4003, line 143: why are we leaving this commented?

hidden4003 (approved):

src/components/Collection/SeriesInfo.tsx
ElementalCrisis: Ah I forgot to delete it as well.

PR #678, opened 2023-11-14
Add Ability to Select Default Poster/Fanart/Banner.

ElementalCrisis:
While you can select a default fanart and it does update, we currently have it set to show a random fanart within a series. We will need to add logic for enabling/disabling random fanart/poster and fix this in the future. 

I also fixed a key issue in series tags.

### DIFF
--- a/src/core/rtkQuery/splitV3Api/seriesApi.ts
+++ b/src/core/rtkQuery/splitV3Api/seriesApi.ts
@@ -200,6 +200,15 @@ const seriesApi = splitV3Api.injectEndpoints({
       }),
     }),
 
+    changeSeriesImage: build.mutation<ResponseType, { seriesId: string, imageType: string, params: object }>({
+      query: ({ imageType, params, seriesId }) => ({
+        url: `Series/${seriesId}/Images/${imageType}`,
+        method: 'PUT',
+        body: params,
+      }),
+      invalidatesTags: ['SeriesAniDB'],
+    }),
+
     // Queue a refresh of all the TvDB data linked to a series using the seriesID.
     refreshSeriesTvdbInfo: build.mutation<boolean, { seriesId: number, force?: boolean }>({
       query: ({ force = false, seriesId }) => ({
@@ -230,6 +239,7 @@ const seriesApi = splitV3Api.injectEndpoints({
 });
 
 export const {
+  useChangeSeriesImageMutation,
   useDeleteSeriesMutation,
   useGetAniDBRecommendedAnimeQuery,
   useGetAniDBRelatedQuery,
--- a/src/core/rtkQuery/splitV3Api/seriesApi.ts
+++ b/src/core/rtkQuery/splitV3Api/seriesApi.ts
@@ -198,15 +198,19 @@ const seriesApi = splitV3Api.injectEndpoints({
       query: ({ seriesId }) => ({
         url: `Series/${seriesId}/Images`,
       }),
+      providesTags: ['SeriesImages'],
     }),
 
-    changeSeriesImage: build.mutation<ResponseType, { seriesId: string, imageType: string, params: object }>({
-      query: ({ imageType, params, seriesId }) => ({
-        url: `Series/${seriesId}/Images/${imageType}`,
+    changeSeriesImage: build.mutation<
+      ResponseType,
+      { seriesId: string, image: Pick<ImageType, 'ID' | 'Source' | 'Type'> }
+    >({
+      query: ({ image, seriesId }) => ({
+        url: `Series/${seriesId}/Images/${image.Type}`,
         method: 'PUT',
-        body: params,
+        body: { ID: image.ID, Source: image.Source },
       }),
-      invalidatesTags: ['SeriesAniDB'],
+      invalidatesTags: ['SeriesAniDB', 'SeriesImages'],
     }),
 
     // Queue a refresh of all the TvDB data linked to a series using the seriesID.
--- a/src/core/rtkQuery/splitV3Api/seriesApi.ts
+++ b/src/core/rtkQuery/splitV3Api/seriesApi.ts
@@ -198,7 +198,7 @@ const seriesApi = splitV3Api.injectEndpoints({
       query: ({ seriesId }) => ({
         url: `Series/${seriesId}/Images`,
       }),
-      providesTags: ['SeriesImages'],
+      providesTags: ['SeriesAniDB'],
     }),
 
     changeSeriesImage: build.mutation<
@@ -210,7 +210,7 @@ const seriesApi = splitV3Api.injectEndpoints({
         method: 'PUT',
         body: { ID: image.ID, Source: image.Source },
       }),
-      invalidatesTags: ['SeriesAniDB', 'SeriesImages'],
+      invalidatesTags: ['SeriesAniDB'],
     }),
 
     // Queue a refresh of all the TvDB data linked to a series using the seriesID.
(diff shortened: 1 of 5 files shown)


src/pages/collection/series/SeriesTags.tsx
```diff
@@ -74,7 +74,7 @@ const SeriesTags = () => {
           </div>
         </div>
         <div className="grid grid-cols-3 gap-8">
-          {map(tags, item => <SeriesTag item={item} />)}
+          {map(tags, (item, index) => <SeriesTag key={index} item={item} />)}
```
hidden4003, line 77: It is generally advised to use the actual id if it is available rather than an array index. Check the pitfall note here https://react.dev/learn/rendering-lists#where-to-get-your-key

hidden4003 (changes requested):

src/pages/collection/series/SeriesTags.tsx
ElementalCrisis: Fixed.

src/core/rtkQuery/splitV3Api/seriesApi.ts
```diff
@@ -200,6 +200,15 @@ const seriesApi = splitV3Api.injectEndpoints({
       }),
     }),
 
+    changeSeriesImage: build.mutation<ResponseType, { seriesId: string, imageType: string, params: object }>({
```
revam, line 203: Suggestion; use `image: Pick<ImageType, 'ID' | 'Source' |'Type">` instead of `imageType: string, params: object`, then map the properties of `image` where they are needed. This will make it so we can just give the image object to the mutation method when using the mutation.

src/core/rtkQuery/splitV3Api/seriesApi.ts
```diff
@@ -200,6 +200,15 @@ const seriesApi = splitV3Api.injectEndpoints({
...
+      query: ({ imageType, params, seriesId }) => ({
+        url: `Series/${seriesId}/Images/${imageType}`,
+        method: 'PUT',
+        body: params,
+      }),
+      invalidatesTags: ['SeriesAniDB'],
+    }),
```
revam, line 210: We need a new tag for `'SeriesImages'` to invalidate, or to just reuse the existing `'SeriesAniAD'` tag for the image endpoint so we can refresh the image data.

src/pages/collection/series/SeriesImages.tsx
```diff
@@ -96,7 +97,16 @@ const SeriesImages = () => {
           <Button
             buttonType="primary"
             className="rounded-md border border-panel-border px-4 py-3 font-semibold"
-            disabled
+            disabled={!Object.keys(selectedImage).length}
```
revam, line 100: Also, we should also probably disable the button if it's already the default/preferred image. That, or turn it into a 'Remove As Default Series Poster'

src/pages/collection/series/SeriesImages.tsx
```diff
@@ -96,7 +97,16 @@ const SeriesImages = () => {
...
+              changeImage({
+                seriesId,
+                imageType: selectedImage.Type,
+                params: { ID: selectedImage.ID, Source: selectedImage.Source },
+              })
+                .then(() => toast.success(`Series ${selectedImage.Type} has been changed.`))
+                .catch(console.error);
+            }}
           >
             Set As Series Poster
```
revam, line 111: The 'Poster' part should be dynamic based on the currently selected image view imo.

revam (commented):
A few comments. I'll leave the rest to @hidden4003 and @ElementalCrisis to decide what to do with.

src/pages/collection/series/SeriesImages.tsx
harshithmohan: I agree

src/pages/collection/series/SeriesImages.tsx
harshithmohan: Either what revam said or just make it "Set as default"

revam: That also works. 👍

src/pages/collection/series/SeriesImages.tsx
ElementalCrisis: Disabling the button makes more sense than changing the text/option to remove.

src/core/rtkQuery/splitV3Api/seriesApi.ts
ElementalCrisis: I added this so the series poster would change when a new default is selected since the series poster already is using ``SeriesAniDB``, but I can change it.

hidden4003: Why do we need a new tag? We don't use separate image tag anywhere as far as I can tell.

revam: The tag would need to be added. It's not currently used since we don't have any mutations that modify the image data. it's also neither hooked up to update when the series data is updated.

src/pages/collection/series/SeriesImages.tsx
ElementalCrisis: > > Disabling the button makes more sense than changing the text/option to remove.
> 
> How do we reset it back to default then?

Honestly, I don't see a reason why we need an option to change back to the default. If you're unhappy with the current image, you can just change it to whichever makes you happy.

da3dsoul: Unmarking makes sense for the cases where a random poster is returned. Those will always return the default if one is set, instead of random

src/core/rtkQuery/splitV3Api/seriesApi.ts
ElementalCrisis: Thanks for the tip!

src/core/rtkQuery/splitV3Api/seriesApi.ts
ElementalCrisis: I added a new one for images and use that one now.

src/pages/collection/series/SeriesImages.tsx
ElementalCrisis: IMO, this should work fine. 

https://github.com/ShokoAnime/Shoko-WebUI/assets/9443295/b49bceec-2602-4bcb-b962-e222921017b7

src/core/rtkQuery/splitV3Api/seriesApi.ts
hidden4003: But that tag `SeriesImages` makes no sense, for example you have  images fetched for multiple series in cache, then by invalidating this tag you will be resetting all the other series images as well because your `SeriesImages` tag is not specific enough, it does not have series id in it. See rtkq docs here; https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#invalidating-tags.

`SeriesAniDB` tag it is also potentially has the issue.

src/pages/collection/series/SeriesImages.tsx
harshithmohan: How does that fix what da3d mentioned?

ElementalCrisis: Lets clarify something here.
 
Currently, our random poster implementation does not update the default poster. So, a group, which normally shows the main series poster will have a random poster selected and when you go inside that group and depending on the number of posters you have, it will also be different. However, inside the series, it always shows the default poster. 

I'm not here to fix this issue in this PR, maybe in a future PR or someone else can. I am saying however that we do not need a default button because its simply not needed,  A default button makes sense on the dashboard for widgets since you can make some drastic changes as it would take too long to manually re-arrange the dashboard. Here you just need to click the first image and then click the button.

hidden4003 (approved):

PR #680, opened 2023-11-14
Misc. changes

revam:
- `misc`: Fix up method signatures.

- `misc`: Add more fields to collection group type.

- `feat`: Add proper redirect when unauthenticated.

- `misc`: Add a redirect when clicking on the series name in the top-right corner on the login screen, for easier removal of incorrect TvDB/TDMB auto-links.

- `misc`: Display the episode type on the card, making it easier to distinguish which type it is if you have multiple types in view, or just in general.

- `misc`: Remove the duplicate 'files' ending for the 'ignored files' tab in the unrecognized utility.

  It would previously read;
  "Unrecognized Files > Unrecognized | Manually Linked | Ignored Files"
  which have now been changed to;
  "Unrecognized Files > Unrecognized | Manually Linked | Ignored"

- `misc`: Remove empty files.

- `fix`: Don't try to refresh and discard files query before returning from the manually linking page to the unrecognized files utility. This was slowing the process down **a lot** while I was trying to mass manually link files. The files query on the unrecognized utility
  also might require different parameters the next time it renders so the response we got will be discarded anyways. :shrug:

- `misc`: Update the 'Add to AniDB' modal (previously known as the 'Copy ED2K Hashes' modal) to be more useful and less finicky.

  This adds a new button at the bottom to re-scan the files, fixes the stall
  search results from a previous search or use of the modal, and disables
  the copy button if no files are selected.

- `feat`: Add ability to remove TvDB links, and also make it display all links for a given site, since you want to delete the right link if you have multiple.

- `fix`: Fix missing stats for series.

- `misc`: Update series info to include the restricted flag and fix the air date and status if the start date is unknown.
  E.g.

  image 
  image

- `fix`: Fix error rendering series page while data is fetching.

- `fix`: Fix navigation to login page while logged in not redirecting the user away from the login page.

- `fix`: Fix row selection on unrecognized table by changing the id selector for each row.

  **FINALLY!**

  No more shifting around the selection!

- `feat`: Add the ability to delete files for episodes on the series page using a confirm modal. Because what kind of monster have a single button to delete files without confirmation?

- `misc`: rename 'File Count' → 'Local' for better clarity of what it is. it is not files, and the count part is infer-able. it is for _local_ episodes.

- `fix`: Fix up the delete button so it matches the mock-up.

- `fix`: Fix AniDB Release Group link for files, and apply a band-aid to stop re-fetching the same data over and over… but only for series data.

- `fix`: Move restricted indicator from the series info panel to a in indicator on the poster per the new mock-up.

- `fix`: Move around the series info panels to align with the mock-up.

- `fix`: Fix up "Add Series to AniDB" modal to align more with the mock-up.

- `fix`: Fix "1 Files" when displayed number of files equals 1.

### DIFF
--- a/src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
+++ b/src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { mdiFileDocumentMultipleOutline, mdiLoading, mdiMagnify, mdiOpenInNew } from '@mdi/js';
 import { Icon } from '@mdi/react';
 import { debounce, forEach } from 'lodash';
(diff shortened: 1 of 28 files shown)


src/hooks/query.ts
```diff
@@ -0,0 +1,116 @@
+import { useCallback, useEffect, useState } from 'react';
```
revam, line 1: I didn't trim it down to just the `useHashQuery` function since the others might be useful other places in the future. Do say if I should had trimmed it down instead. Also, I could had changed the usage to a `useHashQueryParameter` instead for the `LoginPage.tsx` now that I think about it.

revam:
I'm done manually linking now, so i don't think I will be adding any more commits to this PR.

revam:
I lied. I still have one more commit to push. :pensive:

ElementalCrisis:
Couple things.

Already discussed nested grouping, it needs more input and workshopping before we can add it. 

Since you added an item to the series info panel, you need to now increase the dimensions of the series poster to remove the gap at the bottom of the series main panel. However, I think we need to remove the restricted line since it will not apply for 99% of series and instead show an indicator for "Adult" somewhere else. Not sure where it should go instead, current idea is a banner on the series poster. 

Delete Files should be a button and not an option as the toolbar as we want it to stick out, I updated the example in v7 as the episodes section was completed so I didn't bring it over. 

image

Looks like you changed the code for missing which broke the check so now every series will show missing episodes and specials regardless if true. 

image

I don't think we should allow the user to remove the TVDB link as there's currently no way to add it back via the Web UI.

revam:
1. And i said i will remove the commit in question.
2. Fyi, around 10% of all of anidb is restricted content. I can remove it for now, but please decide where to put the indicator soon-ish.
4. Is your server up to date? It works flawlessly for me on my up to date server.
5. There likely won't ever be a way to add it. Unless you want to invest time into adding proper tvdb support in the web ui before we remove it again. Should i still remove it?

da3dsoul:
An 18+ icon on the poster sounds like a good idea. It's also going to have an 18 restricted tag in the list, and most hentai...is obvious by the poster

ElementalCrisis:
We can do something like this.

image

harshithmohan:
image

revam:
> We can do something like this.

Good enough.

src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -53,30 +64,56 @@ function AvDumpSeriesSelectModal({ getLinks, onClose, show }: Props) {
     handleSearch(commonSeries);
   }, [commonSeries, handleSearch]);
 
+  useLayoutEffect(() => () => {
```
hidden4003, line 67: why are we using useLayoutEffect react docs recommend using useEffect

revam: Because the _effect_ needs to happen before the DOM paint, otherwise there will be visual glitches. I've tested it with both `useEffect` and `useLayoutEffect`, and when I used the former, then the visual glitches still remained.

harshithmohan: why is this `() => () => {`?
On line 59 too

revam: Because i want a cleanup effect. If you have a better way to write it that doesn't inteoduce the visual glitching again then please come with suggestions.

harshithmohan: Right... cleanup. I got confused because there was no actual effect, only cleanup. Have never written it that way

PR #681, opened 2023-11-17
Add dashboard settings

harshithmohan:


harshithmohan:
Current dependencies on/for this PR:
* master
  * **PR #681** <a href="https://app.graphite.dev/github/pr/ShokoAnime/Shoko-WebUI/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ShokoAnime/Shoko-WebUI/681?utm_source=stack-comment).

src/core/types/api.ts
```diff
@@ -21,6 +21,11 @@ export type PaginationType = {
   page?: number;
 };
 
+export type DashboardPaginationType = PaginationType & {
+  includeRestricted: boolean;
+  onlyUnwatched?: boolean;
```
revam, line 26: `onlyUnwatched` is only for the Next-Up endpoint. Why is it included in this type when it's not used by the other endpoints? :thinking::thought_balloon:

src/pages/dashboard/DashboardPage.tsx
```diff
@@ -115,30 +132,42 @@ function DashboardPage() {
...
-      <div key="queueProcessor">
-        <QueueProcessor />
-      </div>
+      {!hideQueueProcessor && (
+        <div key="queueProcessor">
+          <QueueProcessor />
+        </div>
+      )}
       <div key="importBreakdown">
         <UnrecognizedFiles />
```
revam, line 141: Why can't we hide the unrecognized files? @ElementalCrisis @harshithmohan

src/pages/dashboard/DashboardPage.tsx
```diff
@@ -115,30 +132,42 @@ function DashboardPage() {
...
+          <RecentlyImported />
+        </div>
+      )}
+      {!hideCollectionBreakdown && (
+        <div key="collectionBreakdown">
+          <CollectionStats />
+        </div>
+      )}
       <div key="collectionTypeBreakdown">
         <MediaType />
```
revam, line 154: Same as above.

src/pages/dashboard/DashboardPage.tsx
```diff
@@ -115,30 +132,42 @@ function DashboardPage() {
...
+          <ShokoNews />
+        </div>
+      )}
+      {!combineContinueWatching && (
+        <div key="continueWatching">
+          <ContinueWatching />
+        </div>
+      )}
       <div key="nextUp">
         <NextUp />
```
revam, line 172: Same as above, but also toggling hiding the next up and continue watching sections separately, and not just combining them.

src/components/Dashboard/DashboardSettingsModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+      tempSettings.WebUI_Settings.dashboard[key] = value as boolean;
+    }
+    setNewSettings(tempSettings);
+  };
+
+  const handleSave = async () => {
+    try {
+      await patchSettings({ oldSettings: settings, newSettings }).unwrap();
+      onClose();
+    } catch (error) { /* empty */ }
```
revam, line 80: We should _at least_ log the error, and maybe handle the error. The user might start wondering _why_ it won't save properly without any indicators.

revam (commented):
I looked through it.

src/pages/dashboard/DashboardPage.tsx
ElementalCrisis: I added the option to hide it.

src/pages/dashboard/DashboardPage.tsx
ElementalCrisis: I added the option to hide it.

ElementalCrisis:
Got bored so added the missing options for the widgets. I also changed the display of the modal but not 100% on the design yet.

src/core/types/api.ts
harshithmohan: That's why it's optional.

src/components/Dashboard/DashboardSettingsModal.tsx
harshithmohan: The error is shown in the toast notification already if it fails

harshithmohan:
@ElementalCrisis I'll give you some work to do 🙃 
Normalize the widths for all modals so we can specify `size` instead of arbitrary widths

src/components/Input/Checkbox.tsx
```diff
@@ -27,13 +30,14 @@ function Checkbox({ className, id, intermediate, isChecked, justify, label, labe
...
         focused && 'ring-2 ring-panel-icon-action ring-inset',
+        disabled && 'opacity-50 cursor-auto',
       ])}
     >
       <input
         id={id}
         type="checkbox"
-        checked={isChecked}
-        onChange={onChange}
+        checked={!disabled && isChecked}
```
harshithmohan, line 39: There may be a possibility of a checked checkbox being disabled somewhere.

src/pages/dashboard/panels/UpcomingAnime.tsx
```diff
@@ -1,20 +1,27 @@
...
 
 const UpcomingAnime = () => {
   const layoutEditMode = useSelector((state: RootState) => state.mainpage.layoutEditMode);
 
   const [showAll, setShowAll] = useState(false);
-  const localItems = useGetDashboardAniDBCalendarQuery({ showAll: false });
-  const items = useGetDashboardAniDBCalendarQuery({ showAll: true });
+
+  const settingsQuery = useGetSettingsQuery();
+  const settings = useMemo(() => settingsQuery.data ?? initialSettings, [settingsQuery.data]);
```
harshithmohan, line 20: Dependency array should be `[settingsQuery]`.
If you use `settingsQuery.data`, useMemo will be able to do nothing and it will change value on every render

harshithmohan: Should ideally be `settingsQuery.requestId` but linter will complain

ElementalCrisis:
> @ElementalCrisis I'll give you some work to do 🙃 Normalize the widths for all modals so we can specify `size` instead of arbitrary widths

Its an issue with the current design. I need to work on it some more.

src/components/Input/Checkbox.tsx
ElementalCrisis: Only if the ``disabled`` prop is changed to true since its optional and always has an initial value of false.

PR #688, opened 2023-11-21
Fix shift select

revam:
Fixes the shift-click I broke when I changed how we choose the row id.

Also added the commit to fix the spot i missed in the last batch of changes… because I don't want to do another PR for that.

### DIFF
--- a/src/components/Dialogs/DeleteFilesModal.tsx
+++ b/src/components/Dialogs/DeleteFilesModal.tsx
@@ -21,7 +21,8 @@ const Title = ({ fileCount }: { fileCount: number }) => (
     <div>Delete Confirmation</div>
     <div className="flex gap-x-1">
       <div className="text-panel-text-important">{fileCount}</div>
-      &nbsp;Files
+      &nbsp;
+      {fileCount === 1 ? 'File' : 'Files'}
     </div>
   </div>
 );
--- a/src/components/Utilities/UtilitiesTable.tsx
+++ b/src/components/Utilities/UtilitiesTable.tsx
@@ -25,15 +25,7 @@ const criteriaMap = {
   created: FileSortCriteriaEnum.CreatedAt,
 };
 
-function isFileType(target: FileType | SeriesType): target is FileType {
-  // @ts-expect-error - Trying to access an unknown property.
-  return typeof target.ID === 'number';
-}
-
-function selectRowId(target: FileType | SeriesType): string | number {
-  return isFileType(target) ? target.ID : target.IDs.ID;
-}
-
+const selectRowId = (target: FileType | SeriesType) => 'ID' in target ? target.ID : target.IDs.ID;
 function UtilitiesTable(props: Props) {
   const {
     setSortCriteria,
--- a/src/components/Utilities/UtilitiesTable.tsx
+++ b/src/components/Utilities/UtilitiesTable.tsx
@@ -25,6 +25,15 @@ const criteriaMap = {
   created: FileSortCriteriaEnum.CreatedAt,
 };
 
+function isFileType(target: FileType | SeriesType): target is FileType {
+  // @ts-expect-error - Trying to access an unknown property.
+  return typeof target.ID === 'number';
+}
+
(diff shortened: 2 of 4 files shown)


src/components/Utilities/UtilitiesTable.tsx
```diff
@@ -25,6 +25,15 @@ const criteriaMap = {
   created: FileSortCriteriaEnum.CreatedAt,
 };
 
+function isFileType(target: FileType | SeriesType): target is FileType {
```
harshithmohan, line 28: so `target` shouldn't be `SeriesType`??

src/components/Utilities/UtilitiesTable.tsx
```diff
@@ -25,6 +25,15 @@ const criteriaMap = {
...
+  // @ts-expect-error - Trying to access an unknown property.
+  return typeof target.ID === 'number';
+}
+
+function selectRowId(target: FileType | SeriesType): string | number {
+  return isFileType(target) ? target.ID : target.IDs.ID;
```
harshithmohan, line 34: why is `isFileType` a separate function when you could've just written that condition here? Also the type doesn't need to be checked. `target.ID ?? target.IDs.ID` would be fine.

harshithmohan (approved):

src/components/Utilities/UtilitiesTable.tsx
```diff
@@ -57,9 +66,11 @@ function UtilitiesTable(props: Props) {
       const lrIndex = lastRowSelected?.current?.index ?? row.index;
       const fromIndex = Math.min(lrIndex, row.index);
       const toIndex = Math.max(lrIndex, row.index);
+      const isSelected = lastRowSelected.current?.getIsSelected() ?? true;
```
harshithmohan, line 69: This is not a complaint to you but to myself. I don't know why I had that inside the loop 🤦

harshithmohan:
Current dependencies on/for this PR:
* master
  * **PR #688** <a href="https://app.graphite.dev/github/pr/ShokoAnime/Shoko-WebUI/688?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #689** <a href="https://app.graphite.dev/github/pr/ShokoAnime/Shoko-WebUI/689?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ShokoAnime/Shoko-WebUI/688?utm_source=stack-comment).

src/components/Dialogs/DeleteFilesModal.tsx
```diff
@@ -21,7 +21,8 @@ const Title = ({ fileCount }: { fileCount: number }) => (
     <div>Delete Confirmation</div>
     <div className="flex gap-x-1">
       <div className="text-panel-text-important">{fileCount}</div>
-      &nbsp;Files
+      &nbsp;
```
hidden4003, line 24: why do we have `&nbsp;` and not a padding? Isn't `&nbsp;` inconsistent between systems and fonts?

hidden4003 (approved):

src/components/Dialogs/DeleteFilesModal.tsx
harshithmohan: `&nbsp;` is basically just a normal space. We write that because we can't put a "space" there. It cannot be padding because it has to follow the font/system's default word spacing

src/components/Utilities/UtilitiesTable.tsx
revam: Try it and find out why tsc and eslint doesn't like it. 😃

revam: The `isFileType` is a type guard that ensures the target is a `FileType`. So tsc is happy since we narrowed down the type.

harshithmohan (changes requested):

PR #690, opened 2023-11-21
Add `Mark filtered as watched` functionality

harshithmohan:


src/core/rtkQuery/splitV3Api/seriesApi.ts
```diff
@@ -81,6 +85,16 @@ const seriesApi = splitV3Api.injectEndpoints({
       providesTags: ['SeriesEpisodes', 'UtilitiesRefresh'],
     }),
 
+    // Toggles the watched state for all the episodes that fit the query.
```
hidden4003, line 88: Does it toggle or does it set watched to specified value? Because toggle means to switch the status to next one usually, so phrasing is confusing.

hidden4003 (approved):

src/core/rtkQuery/splitV3Api/seriesApi.ts
harshithmohan: I copied the comment from the server 🤷 .
We can change it on our side

revam: It toggles it on or off. Or so my late night brain was probably thinking when it made my fingers write thst comment. 🧠

Issue #696, opened 2023-11-23
Wrong episode prefixes


harshithmohan:
![image](https://github.com/ShokoAnime/Shoko-WebUI/assets/26010946/60c66ab7-104d-4e0f-8d07-b53645186b42)

Currently, we are taking the first character of the `Type` of the episode. In this case, it is 'T' because the type is `ThemeSong`. Ideally, this be 'C' to match AniDB's prefix.

Need to map episode types to prefixes to match AniDB and use those instead

revam:
We can also take it one step further and parse if it is OP1, ED1, etc. So it matches what anidb displays on their site.

harshithmohan:
Fixed in #701

PR #701, opened 2023-11-25
Fix episode wrong prefixes

duehoa1211:
This commit fix for issue #696 on front end
I think we should update the type value of back-end as well to match with front-end EpisodeType

### DIFF
--- a/src/pages/dashboard/components/EpisodeDetails.tsx
+++ b/src/pages/dashboard/components/EpisodeDetails.tsx
@@ -74,13 +74,21 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
       return '';
     }
 
+    // Prefixes for episode types base on https://wiki.anidb.net/Content:Episodes
     switch (type) {
       case EpisodeTypeEnum.OpeningSong:
         return 'OP';
       case EpisodeTypeEnum.EndingSong:
         return 'ED';
       case EpisodeTypeEnum.Special:
+      case EpisodeTypeEnum.Extra:
         return 'S';
+      case EpisodeTypeEnum.Trailer:
+        return 'T';
+      case EpisodeTypeEnum.Other:
+        return 'O';
+      case EpisodeTypeEnum.Parody:
+        return 'P';
       default:
         return '';
     }
--- a/src/pages/dashboard/components/EpisodeDetails.tsx
+++ b/src/pages/dashboard/components/EpisodeDetails.tsx
@@ -58,6 +58,28 @@ const TitleSection: React.FC<{ episode: DashboardEpisodeDetailsType, title: stri
   </>
 );
 
+const anidbEpisodePrefixes = (type: EpisodeTypeEnum, epNumber: number): string => {
+  const fullPrefixes = (prefix: string) => `${prefix}${epNumber}`;
+  // Prefixes for episode types base on https://wiki.anidb.net/Content:Episodes#Type
+  switch (type) {
+    case EpisodeTypeEnum.ThemeSong:
+    case EpisodeTypeEnum.OpeningSong:
+    case EpisodeTypeEnum.EndingSong:
+      return fullPrefixes('C');
+    case EpisodeTypeEnum.Special:
+    case EpisodeTypeEnum.Extra:
+      return fullPrefixes('S');
+    case EpisodeTypeEnum.Trailer:
+      return fullPrefixes('T');
+    case EpisodeTypeEnum.Other:
+      return fullPrefixes('O');
+    case EpisodeTypeEnum.Parody:
+      return fullPrefixes('P');
+    default:
+      return fullPrefixes('');
+  }
+};
+
 function EpisodeDetails({ episode, isInCollection = false, showDate = false }: Props): React.ReactNode {
   const percentage = useMemo(() => {
     if (episode.ResumePosition == null) return null;
@@ -65,28 +87,6 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
     const resumePosition = dayjs.duration(convertTimeSpanToMs(episode.ResumePosition));
     return `${((resumePosition.asMilliseconds() / duration.asMilliseconds()) * 100).toFixed(2)}%`;
   }, [episode.Duration, episode.ResumePosition]);
-
-  const anidbEpisodePrefixes = (type: EpisodeTypeEnum, epNumber: number): string => {
-    const fullPrefixes = (prefix: string) => `${prefix}${epNumber}`;
-    // Prefixes for episode types base on https://wiki.anidb.net/Content:Episodes#Type
-    switch (type) {
-      case EpisodeTypeEnum.ThemeSong:
-      case EpisodeTypeEnum.OpeningSong:
-      case EpisodeTypeEnum.EndingSong:
-        return fullPrefixes('C');
-      case EpisodeTypeEnum.Special:
-      case EpisodeTypeEnum.Extra:
-        return fullPrefixes('S');
-      case EpisodeTypeEnum.Trailer:
-        return fullPrefixes('T');
-      case EpisodeTypeEnum.Other:
-        return fullPrefixes('O');
-      case EpisodeTypeEnum.Parody:
-        return fullPrefixes('P');
-      default:
-        return fullPrefixes('');
-    }
-  };
 
   const airDate = useMemo(() => dayjs(episode.AirDate), [episode.AirDate]);
   const relativeTime = useMemo(() => airDate.calendar(null, CalendarConfig), [airDate]);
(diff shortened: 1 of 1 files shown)


harshithmohan:
I will let @revam comment

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,36 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
...
+        return '';
+    }
+  };
+
   const airDate = useMemo(() => dayjs(episode.AirDate), [episode.AirDate]);
   const relativeTime = useMemo(() => airDate.calendar(null, CalendarConfig), [airDate]);
   const title = useMemo(
-    () => `${episode.Type === EpisodeTypeEnum.Normal ? '' : episode.Type[0]}${episode.Number} - ${episode.Title}`,
+    () => {
+      if (episode.Type === EpisodeTypeEnum.Normal) {
```
harshithmohan, line 93: Is this `if` statement required? Because if this doesn't exist the result would be the same anyway. `anidbEpisodeAlias` would return an empty string.

harshithmohan:
We should have prefixes for remaining types also such as PV, Others, Trailers etc

src/pages/dashboard/components/EpisodeDetails.tsx
duehoa1211: yeah, i will remove that, that is no needed at all

duehoa1211:
Base on [this link](https://wiki.anidb.net/Content:Episodes#Type) classify for each episode type, I don't see the a set prefixes for Unknown, and Interview types. Should I make it as empty string like a normal episode, or I missing something?

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,38 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
...
+    }
+
+    // Prefixes for episode types base on https://wiki.anidb.net/Content:Episodes
+    switch (type) {
+      case EpisodeTypeEnum.OpeningSong:
+        return 'OP';
+      case EpisodeTypeEnum.EndingSong:
+        return 'ED';
+      case EpisodeTypeEnum.Special:
+      case EpisodeTypeEnum.Extra:
```
revam, line 84: Extra is not used…  yet.

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,38 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
...
+      default:
+        return '';
+    }
+  };
+
   const airDate = useMemo(() => dayjs(episode.AirDate), [episode.AirDate]);
   const relativeTime = useMemo(() => airDate.calendar(null, CalendarConfig), [airDate]);
   const title = useMemo(
-    () => `${episode.Type === EpisodeTypeEnum.Normal ? '' : episode.Type[0]}${episode.Number} - ${episode.Title}`,
+    () => `${anidbEpisodeAlias(episode.Type, episode.Title)}${episode.Number} - ${episode.Title}`,
```
revam, line 100: If we want to do client-side OP/ED parsing w/variations (e.g. `OP1a`) then I recommend passing in the episode number to the alias function and creating the full string from within the function.

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,38 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
     return `${((resumePosition.asMilliseconds() / duration.asMilliseconds()) * 100).toFixed(2)}%`;
   }, [episode.Duration, episode.ResumePosition]);
 
+  const anidbEpisodeAlias = (type: EpisodeTypeEnum, epTitle: string) => {
```
revam, line 69: This function seems like it can be moved outside the `EpisodeDetails` component, as it doesn't directly use any of the variables from it's parent scopes.

revam (changes requested):
OPs/EDs parsing is a difficult subject to do correctly, especially when the metadata we get from the xml dumps from AniDB doesn't contain the same definitions they show on their site (e.g. we get Credits 3 (`C3`), but they show Opening 1a (`OP1a`)).

I don't intend to scare @duehoa1211 away, but these are the things I would like changed, and hope it's okay with them to make the changes. :pleading_face:

Also, feel free to provide counter arguments if anyone disagrees with what I wrote.

ElementalCrisis:
Thanks for the PR! 

Let us know if you're having issues making the suggested changes either on here or Discord. :)

revam:
> Base on [this link](https://wiki.anidb.net/Content:Episodes#Type) classify for each episode type, I don't see the a set prefixes for Unknown, and Interview types. Should I make it as empty string like a normal episode, or I missing something?
> 
> Edit: I will update for prefixes has been classified

Forgot to say, but the episode types we have exposed in the v3 api is not a 1:1 from anidb's types. We have more types in case future providers (or shoko itself) will use them (now or) in the future.

duehoa1211:
> OPs/EDs parsing is a difficult subject to do correctly, especially when the metadata we get from the xml dumps from AniDB doesn't contain the same definitions they show on their site (e.g. we get Credits 3 (`C3`), but they show Opening 1a (`OP1a`)).
> 
> I don't intend to scare @duehoa1211 away, but these are the things I would like changed, and hope it's okay with them to make the changes. 🥺
> 
> Also, feel free to provide counter arguments if anyone disagrees with what I wrote.

No worry, i am always welcoming advices to better myself.

> > Base on [this link](https://wiki.anidb.net/Content:Episodes#Type) classify for each episode type, I don't see the a set prefixes for Unknown, and Interview types. Should I make it as empty string like a normal episode, or I missing something?
> > Edit: I will update for prefixes has been classified
> 
> Forgot to say, but the episode types we have exposed in the v3 api is not a 1:1 from anidb's types. We have more types in case future providers (or shoko itself) will use them (now or) in the future.

I noticed we do have `unknown` and `interview`, which are not define in anidb, do you have any ideas should I put for these types?

and when i downloading love live's collection i did see something like live performance as well, do we intent add that as well in the future?

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,38 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
...
+    // TODO: This is a hack, we need update backend to make the EpisodeTypeEnum make more sense.
```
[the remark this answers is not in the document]
duehoa1211: This reply just to document this commit
>we will return 'C' for ThemeSong prefixes as default and if the title has Opening or Ending i will replace with OP or ED and keep the rest to match with AniDB 
>this logic will be same for the future if we gonna use Opening and Ending type since AniDb not gonna return song title anytime soon

src/pages/dashboard/components/EpisodeDetails.tsx
duehoa1211: I think I just use the prefixes base on AniDb for `Extra` for now

revam:
> I notice we do have `unknown` and `interview`, which are not define in anidb, do you have any ideas should I put for these types?

Don't even map them for now.

> and when i downloading love live's collection i did see something like live performance as well, do we intent add that as well in the future?

Tl;dr: we only support whatever AniDB support for now.

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -66,10 +66,32 @@ function EpisodeDetails({ episode, isInCollection = false, showDate = false }: P
     return `${((resumePosition.asMilliseconds() / duration.asMilliseconds()) * 100).toFixed(2)}%`;
   }, [episode.Duration, episode.ResumePosition]);
 
+  const anidbEpisodePrefixes = (type: EpisodeTypeEnum, epNumber: number): string => {
```
revam, line 69: The last change I will request is to move the function out of `EpisodeDetails`, since it doesn't depend on the parent scope at all.

revam (changes requested):

src/pages/dashboard/components/EpisodeDetails.tsx
duehoa1211: I'm not quite understanding.
Is that you want me to move it out of `EpisodeDetails`'s component, or move it completely out of `EpisodeDetail`'s files, because both of them is count as out of `EpisodeDetails`'s scope?

revam: Just out of the component is fine. If you want to know where we place them in the file, then I will recommend looking at other files for optimal placement. The general rule of thumb is to place them above the main component (and any other sub-components inside the file).

From my understanding of the existing component files, then the order of declarations would be;
1. imports,
2. local types/interfaces,
3. local constants,
4. local helper functions, 
5. local helper components,
6. main component, and
7. export main component.

revam (approved):

Issue #487, opened 2022-12-12
Series/Group Quick Edit


ElementalCrisis:
https://github.com/ShokoAnime/Shoko-WebUI/assets/9443295/2d015f23-f486-419d-9f02-4384911666d4

Modal height should be static, changing size in video due to a limitation of Figma.

revam:
Is it safe to assume that only one of the "series name" or "group name" sections should be present depending on if we're editing a series or group?

ElementalCrisis:
If editing a group then it's just the group name you can change, we can hide the series name input. However, when editing a series, the user can change the group if they'd like. For example, if they use custom grouping.

PR #707, opened 2023-12-03
Update Edit Series Modal

duehoa1211:
https://github.com/ShokoAnime/Shoko-WebUI/assets/11846133/a4c20af7-95be-47fb-9865-b1b5db1a058e

@ElementalCrisis assigns me with this mockup to update the edit series modal

- This modal is missing saving endpoint for series so I make everything is read only, group's tab still not show for same reason with name
- Personal stats' tab is mockup data only so I will hide it as well


duehoa1211:
Link this to #487

harshithmohan:
`Rehash Files` and `Update AniDB Info - Force` options are missing from actions

src/core/rtkQuery/splitV3Api/collectionApi.ts
```diff
@@ -50,6 +50,39 @@ const collectionApi = splitV3Api.injectEndpoints({
         params: { includeEmpty, topLevelOnly },
       }),
     }),
+
+    getGroupInfinites: build.query<
```
harshithmohan, line 54: Shouldn't this be `getGroupInfinite`?

src/components/Collection/Series/EditSeriesTabs/PersonalStats.tsx
```diff
@@ -0,0 +1,130 @@
...
+      total: 10,
+    },
+    specials: {
+      current: 4,
+      total: 10,
+    },
+  };
+
+  const seriesScore = {
+    isGlobale: true,
```
harshithmohan, line 21:
```suggestion
    isGlobal: true,
```

src/components/Input/InputSmall.tsx
```diff
@@ -23,23 +25,28 @@ function InputSmall(props: Props) {
...
+        placeholder={placeholder ?? ''}
+        value={value}
+        onChange={onChange}
+        onKeyUp={onKeyUp}
+        autoFocus={autoFocus}
+        disabled={disabled}
+        autoComplete={autoComplete ?? 'on'}
+      />
+
+      {suffixes && suffixes(props)}
```
harshithmohan, line 48: Why are the props being passed back to the suffixes function?
They are not being used there.

And the suffixes `<div>` can just be added to the component here and we can just pass the number to this component instead of the function

src/components/Input/Input.tsx
```diff
@@ -90,11 +94,27 @@ function Input(props: Props) {
...
+            <div className="absolute right-3 top-1/2 flex -translate-y-1/2 flex-row gap-x-2">
+              {endIcons.map(icon => (
+                <div
+                  key={`input-${icon.icon}`}
+                  onClick={icon.onClick}
+                  className={cx('cursor-pointer text-panel-text', icon.className)}
+                >
+                  <Icon path={icon.icon} size={1} />
+                </div>
+              ), [] as React.JSX.Element[])}
```
harshithmohan, line 115:
```suggestion
              ), [] as React.ReactNode[])}
```

src/components/Input/Input.tsx
```diff
@@ -90,11 +94,27 @@ function Input(props: Props) {
...
-              className="absolute right-3 top-1/2 -translate-y-1/2 cursor-pointer text-panel-text"
+              className={cx(
+                'absolute right-3 top-1/2 -translate-y-1/2 cursor-pointer text-panel-text',
+                endIconClassName,
+              )}
             >
               <Icon path={endIcon} size={1} horizontal vertical rotate={180} />
             </div>
           )}
+          {endIcons?.length && (
```
harshithmohan, line 105: `endIcon`,`endIconClassName` and `endIconClick` seem redundant now.
Places which are using `endIcon` can be refactored to use `endIcons` instead

src/components/Collection/Series/EditSeriesTabs/NameTab.tsx
```diff
@@ -41,20 +100,27 @@ const NameTab = ({ seriesId }: Props) => {
...
+          if (!search) {
+            acc.push(renderTitle(title));
+            return acc;
+          }
+          if (title.Name.toLowerCase().includes(search.toLowerCase())) {
+            acc.push(renderTitle(title));
+            return acc;
+          }
+          return acc;
+        }, [] as React.JSX.Element[])}
```
harshithmohan, line 122:
```suggestion
        }, [] as React.ReactNode[])}
```

src/components/Collection/Series/EditSeriesTabs/NameTab.tsx
```diff
@@ -17,10 +21,66 @@ const NameTab = ({ seriesId }: Props) => {
     refetchOnMountOrArgChange: false,
   });
 
+  const [fetchSeries, seriesResults] = useLazyGetSeriesInfiniteQuery();
+
+  const getAniDbSeries = (): SeriesTitleType[] => {
```
harshithmohan, line 26: This does not need to be a function. It can be a variable with `useMemo`

src/components/Collection/Series/EditSeriesTabs/NameTab.tsx
```diff
@@ -17,10 +21,66 @@ const NameTab = ({ seriesId }: Props) => {
...
+    const keys = Object.keys(pages);
+    if (!keys?.length) return [];
+
+    return pages[1];
+  };
+
+  const searchSeries = useMemo(() =>
+    debounce(async () => {
+      await fetchSeries({
+        startsWith: search?.length ? search : undefined,
```
harshithmohan, line 39: Is this condition needed?
Because the `useEffect` has `if (!search) return;`, search will always be non-empty

src/components/Collection/Series/EditSeriesTabs/GroupTab.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [search, setSearch] = useState<string>('');
+  const [nameEditable, setNameEditable] = useState<boolean>(false);
+
+  const groupQuery = useGetSeriesGroupQuery({ seriesId: seriesId.toString(), topLevel: false }, {
+    refetchOnMountOrArgChange: false,
+  });
+
+  const [fetchGroup, groupResults] = useLazyGetGroupInfinitesQuery();
+
+  const getAniDbGroup = (): CollectionGroupType[] => {
```
harshithmohan, line 27: can be `useMemo` and not a function

src/components/Collection/Series/EditSeriesTabs/GroupTab.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const keys = Object.keys(pages);
+    if (!keys?.length) return [];
+
+    return pages[1];
+  };
+
+  const searchGroup = useMemo(() =>
+    debounce(async () => {
+      await fetchGroup({
+        startsWith: search?.length ? search : undefined,
```
harshithmohan, line 40: Condition not required. Check similar comment in `NameTab.tsx`
Add `if (!search) return;` to `useEffect` where this function is called.

src/components/Input/InputSmall.tsx
duehoa1211: I trying to aim for a more dynamic render for suffixes if other dev want to render another kind of component or info as a suffixes of the Input

revam: Then just pass on the suffix component itself, don't pass on a render function.

src/components/Collection/Series/EditSeriesTabs/NameTab.tsx
harshithmohan: Resolved by accident?

src/components/Input/InputSmall.tsx
```diff
@@ -11,6 +11,8 @@ type Props = {
   autoFocus?: boolean;
   disabled?: boolean;
   autoComplete?: string;
+
+  suffixes?: React.ReactNode;
```
harshithmohan, line 15:
```suggestion
  suffixes?: React.ReactNode;
```

harshithmohan (changes requested):

harshithmohan (approved):

PR #708, opened 2023-12-04
Add search to collection

harshithmohan:


ElementalCrisis (approved):

revam (approved):
I haven't tested the code, but it looks fine for a dry read through.

Not sure I like how we need to nest the `And` expressions in the builder instead of just providing a single `And` expression with an array of sub-expressions, but that is unrelated to the PR itself and cannot be fixed on the client side anyways, so it won't hold back the PR from being merged.

👍

src/core/rtkQuery/splitV3Api/filterApi.ts
```diff
@@ -0,0 +1,60 @@
...
+  endpoints: build => ({
+    getFilter: build.query<FilterType, { filterId: string }>({
+      query: ({ filterId }) => ({
+        url: `Filter/${filterId}`,
+        params: {
+          withConditions: true,
+        },
+      }),
+    }),
+    getFilteredGroups: build.query<
```
hidden4003, line 20: Maybe we should have some special suffix or prefix for endpoints that merge data? like Infinite

src/pages/collection/Collection.tsx
```diff
@@ -84,6 +94,8 @@ function Collection() {
   const [showDisplaySettingsModal, setShowDisplaySettingsModal] = useState(false);
   const [groupTotal, setGroupTotal] = useState(0);
   const [timelineSeries, setTimelineSeries] = useState<SeriesType[]>([]);
+  const [search, setSearch] = useState('');
+  const debouncedSearch = useDebounce(search, 200);
```
hidden4003, line 98: should we add some rule to only use hook for debounce? I think we use lodash debounce in other places.

hidden4003 (approved):

PR #710, opened 2023-12-06
Rewrite(?) collection view

harshithmohan:


harshithmohan:
Check for any flickers or breakages. Hopefully everything works, as it should

src/pages/collection/Collection.tsx
```diff
@@ -110,32 +118,72 @@ function Collection() {
     patchSettings({ oldSettings: settings, newSettings }).catch(console.error);
   };
 
-  const toggleFilters = () => {
-    setShowFilterSidebar(!showFilterSidebar);
-  };
+  // Q: Why is { skip: isSeries } not added here?
+  // A: When skip parameter changes while navigating to a group and back, the query is refired even if the data already
+  // exists causing flickers in the UI.
```
hidden4003, line 123: Not related to this, just a thought that flickers happen because there is no proper fetching indication to user and stale data is shown instead and then it gets fetched and refreshes but user does not know that fetching was even going on so it looks like a flicker or a glitch.

hidden4003 (approved):

src/pages/collection/Collection.tsx
harshithmohan: I have tried to fix that with this PR. Performing a search and changing filters should not show a loading indicator

PR #712, opened 2023-12-07
Add includeRestricted to RecommendedAnime Dashboard Panel

natyusha:
- requires api update to function

### DIFF
--- a/src/core/rtkQuery/splitV3Api/seriesApi.ts
+++ b/src/core/rtkQuery/splitV3Api/seriesApi.ts
@@ -6,7 +6,7 @@ import {
 } from '@/core/rtkPaginationUtil';
 import { splitV3Api } from '@/core/rtkQuery/splitV3Api';
 
-import type { InfiniteResultType, ListResultType, PaginationType, RecommendedPaginationType } from '@/core/types/api';
+import type { InfiniteResultType, ListResultType, PaginationType } from '@/core/types/api';
 import type { CollectionGroupType } from '@/core/types/api/collection';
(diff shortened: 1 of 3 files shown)


natyusha:
> uhh.. you forgetting something? 😔

As a matter of fact, I forgot almost everything :joy:

harshithmohan:
image
You didn't update the type..
The build job is supposed to catch that but didn't. It needs to be fixed...

src/core/rtkQuery/splitV3Api/seriesApi.ts
```diff
@@ -149,7 +149,7 @@ const seriesApi = splitV3Api.injectEndpoints({
     }),
 
     // Gets anidb recommendation for the user
-    getAniDBRecommendedAnime: build.query<SeriesRecommendedType[], PaginationType>({
+    getAniDBRecommendedAnime: build.query<SeriesRecommendedType[], RecommendedPaginationType>({
```
harshithmohan, line 152:
```suggestion
    getAniDBRecommendedAnime: build.query<SeriesRecommendedType[], PaginationType & { includeRestricted: boolean; }>({
```

I don't see a point on adding a whole another type if it's not going to be used elsewhere. This should be just fine. Remove the type declaration

harshithmohan (changes requested):

src/core/rtkQuery/splitV3Api/seriesApi.ts
natyusha: Applied the requested changes

harshithmohan (approved):

PR #718, opened 2023-12-19
Replace RTK query with tanstack-query

harshithmohan:


src/components/Collection/SeriesInfo.tsx
```diff
@@ -128,12 +125,12 @@ const SeriesInfo = () => {
                 <div className="font-semibold ">Local</div>
                 <div className="flex flex-row gap-x-1">
                   <span>EP:</span>
-                  <span>{formatThousand(series.Sizes.Local.Episodes)}</span>
-                  {series.Sizes.Local.Specials !== 0 && (
+                  <span>{formatThousand((series?.Sizes.Local.Episodes) ?? 0)}</span>
```
hidden4003, line 128: Looking a lot messier, from the docs it looks like we could use `initialData` to have some dummy object and not have to deal with undefined.

src/core/axios.ts
```diff
@@ -0,0 +1,42 @@
...
+  baseURL: '/api',
+});
+
+export const axiosPlex = axiosDefault.create({
+  baseURL: '/api/plex',
+});
+
+export const axiosExternal = axiosDefault.create();
+
+const addApikeyInterceptor = (config: InternalAxiosRequestConfig) => {
```
hidden4003, line 27: why do you have to set it 2 times?

src/pages/collection/Collection.tsx
```diff
@@ -127,68 +123,67 @@ function Collection() {
...
-      TagLimit: 20,
-    }).then().catch(error => console.error(error));
-  }, [currentPage, fetchGroupExtras, groupsQuery]);
-
-  // This can be inside CollectionView but then defaultPageSize has to be passed to it so same either way
-  // 999 to make it effectively infinite since Group/{id}/Series is not paginated
-  const pageSize = useMemo(() => (groupId ? 999 : defaultPageSize), [groupId]);
+    if (!groupExtrasQuery.isSuccess) return;
+    setGroupExtras(immerState => [...immerState, ...groupExtrasQuery.data]);
+  }, [groupExtrasQuery.data, groupExtrasQuery.isSuccess, setGroupExtras]);
```
hidden4003, line 178: why is 3rd dependency `setGroupExtras`?

hidden4003 (approved):

src/components/Collection/Series/EditSeriesTabs/NameTab.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { mdiCheckUnderlineCircleOutline, mdiCloseCircleOutline, mdiMagnify, mdiPencilCircleOutline } from '@mdi/js';
 import cx from 'classnames';
-import { debounce } from 'lodash';
 
 import Input from '@/components/Input/Input';
-import { useGetSeriesQuery, useLazyGetSeriesInfiniteQuery } from '@/core/rtkQuery/splitV3Api/seriesApi';
+import { useSeriesQuery } from '@/core/react-query/series/queries';
+// import { useLazyGetSeriesInfiniteQuery } from '@/core/rtkQuery/splitV3Api/seriesApi';
```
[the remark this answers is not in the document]
harshithmohan: On line 22, I forgot to put the TODO prefix, doing it now

src/components/Collection/SeriesInfo.tsx
harshithmohan: I'll look into it 🤔

src/core/axios.ts
harshithmohan: ... That's a mistake

src/pages/collection/Collection.tsx
harshithmohan: Because we are using it on line 177.. Linter wants it

hidden4003: That is fine then, I was just thinking maybe I missed something in how those setters work.

src/components/Collection/SeriesInfo.tsx
harshithmohan: It was more about typescript thinking series is null when it actually couldn't be. I fixed it without using initial data. But this initial data might be useful for other places 🙃

revam: It is because you're only check if the seriesId is nit falsely and not if the series is null when rendering the component (I think)

PR #735, opened 2023-12-25
Add plugin:@typescript-eslint/recommended-type-checked

hidden4003:


hidden4003:
@revam can you look at errors in  src\core\utilities\auto-match-logic.ts ?
@harshithmohan can you look at  src\core\axios.ts and  src\components\Utilities\UtilitiesTable.tsx ?
 src\core\patches.ts we can probably delete

PR #743, opened 2023-12-30
Sidebar filter

hidden4003:


src/core/react-query/filter/queries.ts
```diff
@@ -29,6 +29,13 @@ export const useSubFiltersQuery = (filterId: number, enabled = false) =>
     enabled,
   });
 
+export const useFilterExpressionsQuery = (enabled = true) =>
```
harshithmohan, line 32: We don't need the `enabled` param since we are not using it but fine

harshithmohan (approved):

PR #791, opened 2024-01-18
Add Api token management screen

duehoa1211:


src/components/Input/Input.tsx
```diff
@@ -132,7 +132,7 @@ function Input(props: Props) {
             disabled={disabled}
             ref={inputRef}
           />
-          {(endIcons?.length ?? isOverlay) && (
+          {((endIcons?.length ?? 0) > 0 || isOverlay) && (
```
harshithmohan, line 135: Does it not mean the same thing?

src/core/react-query/auth/mutations.ts
```diff
@@ -18,3 +18,24 @@ export const useLoginMutation = () =>
       }));
     },
   });
+
+export const useCreateApiToken = () =>
+  useMutation<string, unknown, string>({
```
harshithmohan, line 23:
```suggestion
  useMutation<string>({
```

No need to specify it twice if it is not being transformed.

src/core/react-query/auth/queries.ts
```diff
@@ -0,0 +1,11 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { axiosV2 } from '@/core/axios';
+
+import type { AuthToken } from '@/core/types/api/authToken';
+
+export const useApiKeyQuery = () =>
+  useQuery<AuthToken[]>({
+    queryKey: ['auth', 'auth'],
```
harshithmohan, line 9:
```suggestion
    queryKey: ['auth', 'apikey'],
```

harshithmohan: Also create and delete mutations should invalidate this query

src/pages/settings/SettingsPage.tsx
```diff
@@ -70,6 +70,18 @@ function SettingsPage() {
     }
   };
 
+  const isShowFooter = useMemo(() => {
+    const currentPath = pathname.split('/').pop();
+    const item = items.find(x => x.path === currentPath);
+    return item?.useDefaultFooter ?? true;
+  }, [pathname]);
```
harshithmohan, line 77: Instead of adding the `useDefaultFooter` property to `items` and then doing find on it, the ones which don't need the footer can be in an array like `['user-management', 'api-keys']` and then you can just check if that array contains `pathname.split('/').pop()`.

Or better yet. Move the footer into its own component and use it inside each settings component wherever required and just keep the outlet here.

harshithmohan (changes requested):
I don't know if I was supposed to review it yet. But I did 😔

harshithmohan:
Also, `eslint-disable-next-line @typescript-eslint/no-floating-promises`
Remove those and just add `.catch(console.error)`

src/components/Input/Input.tsx
duehoa1211: Yeah, that was changed to make myself clear when debugging but forgot to change back

src/core/react-query/auth/mutations.ts
duehoa1211: got this error if it not define like that
image
image

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const { data: tokens, refetch } = useApiKeyQuery();
+
+  const onTokenDelete = useEventCallback(() => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    setIsRefetch(_ => true);
+  });
+
+  useEffect(() => {
+    if (!refetch) return;
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
hidden4003, line 82: Same as in other place add `.catch(error.log)` no need to eslint-disable-next-line

hidden4003: Why is refetch not called directly but in a useEffect?

hidden4003 (changes requested):

src/core/react-query/auth/mutations.ts
harshithmohan: Ohh. I was thinking of the query type syntax. The 2nd `string` here is the input param type for the mutation function

harshithmohan (changes requested):

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (!generatedSucceed || !createdToken) return undefined;
+    setKeyValue(_ => createdToken);
+    setInputDisabled(_ => true);
+    setIsRefetch(_ => true);
+    toast.success('API Generated', 'API Key has been generated!', {
+      autoClose: 3000,
+      toastId: 'api-generated',
+    });
+    toast.warning('Copy Your API Key!', 'You won\'t be able to copy this key anymore once you leave this page!', {
+      autoClose: 99999999999,
```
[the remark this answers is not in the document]
duehoa1211: the reason I put value there is because the icon that render with false condition is not match with figma, but the condition with value is

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  const { data: tokens, refetch } = useApiKeyQuery();
+
+  useEffect(() => {
+    if (!generatedSucceed || !createdToken) return undefined;
+    setKeyValue(createdToken);
+    setInputDisabled(true);
+    return () => {
+      if (!toast.isActive('copy-api-key')) return;
```
hidden4003, line 95: This will bail and potentially leave some toasts open, condition needs to be reversed and return is not needed. Should be:
```
if (toast.isActive('copy-api-key')) {
    toast.dismiss('copy-api-key');
}
```

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,148 @@
...
+        toast.warning('Copy Your API Key!', 'You won\'t be able to copy this key anymore once you leave this page!', {
+          autoClose: 99999999999,
+          toastId: 'copy-api-key',
+        });
+        invalidateQueries(['auth', 'apikey']);
+      },
+    });
+  });
+
+  const { data: tokens, refetch } = useApiKeyQuery();
```
hidden4003, line 88: `refetch` is not used anywhere, it is not needed

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const [keyValue, setKeyValue] = useState('');
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [, copy] = useCopyToClipboard();
+  const onKeyChange = useEventCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    setKeyValue(event.target.value);
+  });
+
+  const {
+    data: createdToken,
+    isError: generatedFailed,
```
hidden4003, line 55: `generatedFailed` is not used

hidden4003 (changes requested):

src/pages/settings/tabs/ApiKeys.tsx
```diff
@@ -0,0 +1,148 @@
...
+        <Input
+          id="key-input"
+          endIcons={inputDisabled
+            ? [{
+              icon: mdiClipboardTextOutline,
+              className: 'text-panel-text-secondary',
+              onClick: handleCopyToClipboard,
+            }]
+            : undefined}
+          className={cx(!generatedSucceed ? 'w-full xl:w-[65%]' : 'w-[100%]')}
```
hidden4003, line 120: You need to put `grow` on it not manipulate the field width and put `shrink-0` on a button, maybe adjust flex-basis of a button if needed

ElementalCrisis: I fixed this in my branch as I've made some other changes to settings. No need to fix @duehoa1211.

hidden4003 (approved):

Issue #809, opened 2024-02-07
Ignored Files Not Showing in WebUI


natyusha:
## VERSION INFORMATION

Server Version: Latest

Desktop Version: Latest

## LOG FILE

Nothing showing in the logs

## DESCRIPTION

When marking a file as ignored via the WebUI or Desktop it will not show up in the WebUI but will show in Desktop.

## STEPS TO REPRODUCE

[Example Video](https://github.com/ShokoAnime/ShokoServer/assets/985941/4324d6c1-78e9-4fc9-9921-5a5d10af8088)

Mark a file as ignored then navigate to /webui/utilities/unrecognized/ignored-files and note that nothing will display there.

PR #810, opened 2024-02-07
#809 Fix Ignored Files

ElementalCrisis:
Not sure if this is the right approach but it does fix the following issues.

- Ignoring a file does not refresh the unrecognized list.
- Ignored files show do not show up.
- Restoring a file does not refreshes the ignored list. 

Additionally, ``include: ['Ignored']`` was required when testing in swagger and it seems the endpoint does not work without it now, maybe @da3dsoul can comment on if this is a bug with the API.

### DIFF
--- a/src/core/react-query/file/mutations.ts
+++ b/src/core/react-query/file/mutations.ts
@@ -43,8 +43,8 @@ export const useDeleteFileLinkMutation = () =>
 
 export const useIgnoreFileMutation = () =>
   useMutation({
-    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore`, { ignore }),
-    onSuccess: () => invalidateQueries(['files', { include_only: ['Ignored'] }]),
+    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore?value=${ignore}`),
+    onSuccess: () => invalidateQueries(['files', { include: ['Ignored'], include_only: ['Ignored'] }]),
   });
 
 export const useLinkOneFileToManyEpisodesMutation = () =>
--- a/src/core/react-query/file/mutations.ts
+++ b/src/core/react-query/file/mutations.ts
@@ -44,7 +44,7 @@ export const useDeleteFileLinkMutation = () =>
 export const useIgnoreFileMutation = () =>
   useMutation({
     mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore?value=${ignore}`),
-    onSuccess: () => invalidateQueries(['files', { include: ['Ignored'], include_only: ['Ignored'] }]),
+    onSuccess: () => invalidateQueries(['files']),
   });
 
 export const useLinkOneFileToManyEpisodesMutation = () =>
--- a/src/core/react-query/file/mutations.ts
+++ b/src/core/react-query/file/mutations.ts
@@ -43,7 +43,8 @@ export const useDeleteFileLinkMutation = () =>
 
 export const useIgnoreFileMutation = () =>
   useMutation({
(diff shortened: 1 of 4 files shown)


hidden4003 (approved):

harshithmohan:
Wait for my review, don't merge

src/core/react-query/file/mutations.ts
```diff
@@ -43,8 +43,8 @@ export const useDeleteFileLinkMutation = () =>
 
 export const useIgnoreFileMutation = () =>
   useMutation({
-    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore`, { ignore }),
-    onSuccess: () => invalidateQueries(['files', { include_only: ['Ignored'] }]),
+    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore?value=${ignore}`),
```
harshithmohan, line 46:
```suggestion
    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) => axios.put(`File/${fileId}/Ignore`, { value: ignore }),
```

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
```diff
@@ -117,6 +117,7 @@ const Menu = (
         const failedCount = countBy(result, 'status').rejected;
         if (failedCount) toast.error(`Error ignoring ${failedCount} files!`);
         if (failedCount !== selectedRows.length) toast.success(`${selectedRows.length} files ignored!`);
+        invalidateQueries(['files']);
```
harshithmohan, line 120: This is not needed now since the invalidation on mutation itself has changed. This is duplicate invalidation

src/pages/utilities/UnrecognizedUtilityTabs/IgnoredFilesTab.tsx
```diff
@@ -96,6 +96,7 @@ function IgnoredFilesTab() {
   const filesQuery = useFilesInfiniteQuery(
     {
       pageSize: 50,
+      include: ['Ignored'],
```
harshithmohan, line 99: This is definitely a server bug because it was working before da3d touched it.

harshithmohan (changes requested):

src/core/react-query/file/mutations.ts
hidden4003: You are suggesting to revert it to the way it didn't work. It needs a get parameter it doesn't work otherwise. Swagger says that value is located in path i.e. in url not a post parameter.

src/pages/utilities/UnrecognizedUtilityTabs/IgnoredFilesTab.tsx
hidden4003: include_only should either add this to include silently or throw an error telling you that you are trying to filter something that will never be in the dataset in the first place

src/core/react-query/file/mutations.ts
harshithmohan:
```suggestion
    mutationFn: ({ fileId, ignore }: IgnoreFileRequestType) =>
      axios.put(`File/${fileId}/Ignore`, undefined, { params: { value: ignore } }),
```

my bad, it should actually be this. I thought it was a `get` and not `put`

src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
ElementalCrisis: Did you try? Without this, unrecognized does not update, you have to refresh.

harshithmohan: yes

https://github.com/ShokoAnime/Shoko-WebUI/assets/26010946/ef10066d-91af-48b6-a465-508b92f50cbd

PR #833, opened 2024-03-19
Update logic so Rehash and Rescan can be used on all unrecognized files.

ElementalCrisis:
Allows the user to rehash or rescan all files without having to select any files.


harshithmohan:
This will not work as we don't have *all* the files. The files endpoint is paginated. This is why the options were removed.

ElementalCrisis:
Closing for reasons stated by Mohan. We'll increase the pageSize limit to allow users to select more items as once.

PR #850, opened 2024-04-03
Show `Type` instead of `Title` for queue items when not running

harshithmohan:


da3dsoul:
You should probably make it switch on IsRunning

harshithmohan:
The problem with changing it based on `IsRunning` is, for instances like this, the Type and Title don't exactly match. So it will look like it changed into some other task
image

da3dsoul:
I can fix those. I think I just forgot to update them

PR #852, opened 2024-04-03
Adds the time an AniDB ban is expected to expire to the modal

fearnlj01:
Following on from my bodged work in the server... The reason should now be clear!

This change should provide an estimate of when an AniDB ban will be tested for expiry in the ban information modal. Screenshot of what it looks like as per the below.

2024-04-03_22-39-49_firefox

Whilst I have tested this and it works fine when everything is set from the initial `AniDB:OnConnected` event, I stress that as I've managed to get myself banned again, I can't definitively confirm that this also works as expected after the WebUI receives an updated ban status.


ElementalCrisis:
How about this for the design? 

image

Its put the info on its own line and makes it stick out a bit more.

fearnlj01:
Seems sensible - I wasn't too happy with the wording I'd originally provided!

I've implemented the requested changes now and _quickly_ tested it and all appears ok.

PR #861, opened 2024-04-10
Truncates anime title on login page as rationally expected

fearnlj01:
Prevents the icon from shrinking & ensures that the series name truncates as logically expected.

With changes applied, hovering over the anime title now looks like the below:
2024-04-10_14-35-05_firefox

_(Please don't make me inspect element to get the full anime title when it's truncated :^) )_


fearnlj01:
And for the sake of comparison... A long title _(Keiken Zumi na Kimi to, Keiken Zero na Ore ga, Otsukiai Suru Hanashi.)_ currently appears like this:
image

PR #865, opened 2024-04-13
Selectively clear localStorage

fearnlj01:
I've been wondering why I had to always keep on re-authorizing swagger given that it's meant to have persistent authorization... realised it's more likely than not due to the entirety of `localStorage` being cleared on WebUI updates, rather than selectively clearing entries. (which, after testing my changes, is confirmed to _probably_ be the cause)

`apiSession` & `theme` appear to be the only keys saved in `localStorage` across the repo. Not too sure if it's intended for the theme to be cleared as well, but I'll keep the behaviour the same for now and leave this as a draft PR.

### DIFF
--- a/src/core/localStorage.ts
+++ b/src/core/localStorage.ts
@@ -21,7 +21,8 @@ export const loadState = (): RootState => {
     }
     const apiSession: unknown = JSON.parse(apiSessionString);
     if (!isApiSession(apiSession)) {
-      globalThis.localStorage.clear();
+      globalThis.localStorage.removeItem('apiSession');
+      globalThis.localStorage.removeItem('theme');
       return {} as RootState;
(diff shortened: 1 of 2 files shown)


fearnlj01:
Another potential avenue to explore if people are willing to humour me, but not as intially suggested, would be leaving _only_ the localStorage key set by swagger untouched, as per the below.
```js
for (let i = 0; i < globalThis.localStorage.length ; i++) {
  const key = globalThis.localStorage.key(i);
  if (key !== "authorized") {
    globalThis.localStorage.removeItem(key);
  }
}
```

harshithmohan:
two words: face palm

We can just clear the `apiSession` key. Theme doesn't need to be cleared.

PR #869, opened 2024-04-15
Series Redesign

ElementalCrisis:
Updating design based on latest mockups from 7.5 Rebuild.


src/core/react-query/group/mutations.ts
```diff
@@ -0,0 +1,56 @@
...
+  MoveSeriesGroupRequestType,
+  PatchGroupRequestType,
+} from '@/core/react-query/group/types';
+
+/**
+ * This file probably needs more work on query invalidation.
+ * Currently, it's naïve and only invalidates the current series query.
+ * It should probably also invalidate the cache for:
+ *  * Any series belonging to the "original" groups of a series.
+ */
```
harshithmohan, line 18: @fearnlj01 let's discuss on discord 😔 
It's a bit tricky because we already invalidate on SeriesUpdated event from signalr. So these invalidations might not be required at all.

src/components/Collection/Files/FilesMissingEpisodes.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <div className="flex max-h-[72vh] flex-col rounded-lg border border-panel-border bg-panel-background-transparent p-6 transition-colors">
+        <div className="sticky top-0 z-[1] flex bg-panel-background-alt">
+          <div className="mb-1 flex grow items-center gap-16 rounded-lg border border-panel-border bg-panel-table-header p-4 text-left font-semibold transition-colors">
+            <div className="w-[12.5rem] text-left">
+              Type
+            </div>
+            <div className="w-[46.875rem] text-left">
+              Title
+            </div>
+            <div className="w-[139px] text-left">
```
harshithmohan, line 68: use `rem`...

src/components/Collection/Tags/TagDetailsModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+        },
+      },
+    },
+    show,
+  );
+  const [seriesData, seriesCount] = useFlattenListResult(seriesDataList);
+
+  if (!isFetchingNextPage && seriesData.length !== seriesCount) {
+    fetchNextPage().catch(() => {});
+  }
```
harshithmohan, line 32: could just make it pageSize: 0 if we are loading everything anyway

harshithmohan (changes requested):

harshithmohan:
Regarding the edit series modal, maybe it's better to use redux instead of prop drilling it all the way. But that can be a separate PR

src/components/Collection/Tags/TagDetailsModal.tsx
fearnlj01: Just to provide justificication for my original approach - I wasn't too sure what the tradeoff might be between doing a potentially large/"slow" network request for users with larger collections vs many smaller repeated requests. I decided to err on the side of caution so went to implement this using a page size of 1. Since everything seemed to remain responsive, I kept it as is, albeit with a more reasonable default page size.

If there's no need for performance concerns on this front then happy days, should make things simpler for sure.

harshithmohan: In that case I would argue that the rows should also be virtualized. But again, in a different PR. Leave this as is

harshithmohan (approved):

src/components/Collection/Files/FilesMissingEpisodes.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <div>
+      {dayjs(episode.AirDate).format('MMMM DD YYYY')}
+    </div>
+  </div>
+);
+
+type Props = {
+  missingEps?: WebuiSeriesFileSummaryMissingEpisodeType[];
+};
+const FileMissingEpisodes = ({ missingEps = [] }: Props) => (
```
hidden4003, line 51: Needs some React.memo

src/components/Collection/Files/FilesSummaryGroup.tsx
```diff
@@ -0,0 +1,154 @@
...
+        </div>
+      </div>
+    </div>
+  );
+};
+
+type Props = {
+  groups?: WebuiSeriesFileSummaryGroupType[];
+};
+const FilesSummaryGroups = ({ groups = [] }: Props) => (
```
hidden4003, line 149: Needs react.memo, also react uses array index as key by default so this is not doing anything, is there no id we can use?

src/components/Collection/SeriesTopPanel.tsx
```diff
@@ -0,0 +1,108 @@
...
+      'text-sm font-semibold flex gap-x-3 items-center border-2 border-panel-tags rounded-lg py-2 px-3 whitespace-nowrap capitalize',
+      type === 'User' ? 'text-panel-icon-important' : 'text-panel-icon-action',
+    )}
+  >
+    <Icon path={mdiTagTextOutline} size="1.25rem" />
+    <span className="text-panel-text">{text}</span>
+  </div>
+);
+
+const SeriesTopPanel = ({ series }: SeriesSidePanelProps) => {
```
hidden4003, line 34: Needs react.memo

src/components/Collection/SeriesUserStats.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+import { formatThousand } from '@/core/util';
+
+import type { SeriesType } from '@/core/types/api/series';
+
+type SeriesInfoProps = {
+  series: SeriesType;
+};
+
+const SeriesUserStats = ({ series }: SeriesInfoProps) => (
```
hidden4003, line 11: Needs react.memo

src/components/Input/MultiStateButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type Props = {
+  states: { label?: string, value: string }[];
+  className?: string;
+  alternateColor?: boolean;
+  activeState: string;
+  onStateChange: (state: string) => void;
+};
+
+const MultiStateButton = ({ activeState, alternateColor, className, onStateChange, states }: Props) => (
```
hidden4003, line 14: Needs react.memo

hidden4003:
> Regarding the edit series modal, maybe it's better to use redux instead of prop drilling it all the way. But that can be a separate PR

I was thinking outlet context, but redux works too

harshithmohan:
> > Regarding the edit series modal, maybe it's better to use redux instead of prop drilling it all the way. But that can be a separate PR
> 
> I was thinking outlet context, but redux works too

Series page is not a child of Collection so outlet context wouldn't work

src/components/Collection/Files/FilesSummaryGroup.tsx
harshithmohan: The grouping is dynamic so there's nothing unique in them to add an ID as

PR #921, opened 2024-05-28
Use redux store for edit series modal

fearnlj01:
As TODO'd in PR #869

### DIFF
--- a/src/components/Collection/Series/EditSeriesModal.tsx
+++ b/src/components/Collection/Series/EditSeriesModal.tsx
@@ -26,7 +26,8 @@ const tabs = {
   // stats: 'Personal Stats',
 };
 
-const renderTab = (activeTab: string, seriesId: number) => {
+const renderTab = (activeTab: string, seriesId = -1) => {
+  if (seriesId === -1) return null;
   switch (activeTab) {
     case 'name':
       return <NameTab seriesId={seriesId} />;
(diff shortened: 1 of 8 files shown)


fearnlj01:
I was going to have the setSeriesId action also set the open status of the modal at the same time... But that wasn't how it was done in the `browseFolder` modal, so I've kept things consistent instead 👍

harshithmohan (changes requested):
We don't need to be consistent with browseFolderModal. That has a show parameter because it has nothing else. Here `show` can be just `seriesId !== -1`. We can make do with a single value instead of `show` and `seriesId`

src/components/Collection/Series/EditSeriesModal.tsx
```diff
@@ -48,18 +48,17 @@ const EditSeriesModal = () => {
...
 
   const [activeTab, setActiveTab] = useState('name');
 
   return (seriesId === -1)
     ? null
     : (
-      <ModalPanel show={show} onRequestClose={onClose} header="Edit Series" size="md" noPadding noGap>
+      <ModalPanel show={seriesId !== -1} onRequestClose={onClose} header="Edit Series" size="md" noPadding noGap>
         <div className="flex h-[26rem] flex-row gap-x-6 p-6">
           <div className="flex shrink-0 gap-y-6 font-semibold">
```
harshithmohan: Remove the `(seriesId === -1) ? null :` part otherwise modal transition won't work

harshithmohan (changes requested):

src/components/Collection/Series/EditSeriesModal.tsx
fearnlj01: Went mad trying to work out which transition doesn't work as it looked like it was... Took a dozen mini-recordings for me to realise it was the exit transition 🙃

Fixed in the latest commit 👍

harshithmohan (approved):

PR #923, opened 2024-05-29
Feat: Virtualize series list in tag modal

fearnlj01:
Completes another TODO from PR #869

### DIFF
--- a/src/components/Collection/Tags/TagDetailsModal.tsx
+++ b/src/components/Collection/Tags/TagDetailsModal.tsx
@@ -41,6 +41,10 @@ const SeriesVirtualizer = (
   const virtualizer = useVirtualizer({
     count: dataSize,
     getScrollElement: () => scrollRef.current,
+    getItemKey: (idx) => {
+      const seriesId = data[idx]?.IDs.ID;
+      return seriesId ? `tag-series-id-${seriesId}` : `tag-series-idx-${idx}`;
+    },
     estimateSize: () => (25.6), // Standard line height
     gap: 8,
     overscan: 12,
@@ -49,11 +53,11 @@ const SeriesVirtualizer = (
   return (
     <div className="shoko-scrollbar max-h-[12.5rem] overflow-y-auto" ref={scrollRef}>
       <div className="relative" style={{ height: virtualizer.getTotalSize() }}>
-        {virtualizer.getVirtualItems().map(({ index, size, start }) => {
+        {virtualizer.getVirtualItems().map(({ index, key, size, start }) => {
           const series = data?.[index] ?? fetchNext();
           return (
             <div
-              key={index}
+              key={key}
               className="absolute left-0 top-0 w-full"
               style={{ height: size, transform: `translateY(${start}px)` }}
             >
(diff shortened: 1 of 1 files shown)


harshithmohan (changes requested):
This still loads all pages before scrolling down. It should be on infinite scroll

harshithmohan (changes requested):
image

Well.. _something_ is wrong

fearnlj01:
> image
> 
> Well.. _something_ is wrong

I hadn't considered the possibility of a Shoko Series ID possibly being the same as the current index... The issue with duplicate keys should now be resolved via setting distinctly different keys based on if it's got a series ID as the key, or it's falling back to the index.

I can no longer reproduce it as an issue on my end anymore at the very least! Apologies for letting this one slip through.

src/components/Collection/Tags/TagDetailsModal.tsx
```diff
@@ -1,20 +1,81 @@
...
+  const scrollRef = useRef<HTMLDivElement>(null);
+  const virtualizer = useVirtualizer({
+    count: dataSize,
+    getScrollElement: () => scrollRef.current,
+    getItemKey: (idx) => {
+      const seriesId = data[idx]?.IDs.ID;
+      return seriesId ? `tag-series-id-${seriesId}` : `tag-series-idx-${idx}`;
+    },
+    estimateSize: () => (25.6), // Standard line height
+    gap: 8,
```
harshithmohan, line 49: Why did I not know that this existed? 😔

src/components/Collection/Tags/TagDetailsModal.tsx
```diff
@@ -1,20 +1,81 @@
...
+  { data, dataSize, fetchNext }: { data: SeriesType[], dataSize: number, fetchNext: () => void },
+) => {
+  const scrollRef = useRef<HTMLDivElement>(null);
+  const virtualizer = useVirtualizer({
+    count: dataSize,
+    getScrollElement: () => scrollRef.current,
+    getItemKey: (idx) => {
+      const seriesId = data[idx]?.IDs.ID;
+      return seriesId ? `tag-series-id-${seriesId}` : `tag-series-idx-${idx}`;
+    },
```
harshithmohan, line 47: What happens if we don't have this at all? 🤔

src/components/Collection/Tags/TagDetailsModal.tsx
```diff
@@ -1,20 +1,81 @@
...
+    estimateSize: () => (25.6), // Standard line height
+    gap: 8,
+    overscan: 12,
+  });
+
+  return (
+    <div className="shoko-scrollbar max-h-[12.5rem] overflow-y-auto" ref={scrollRef}>
+      <div className="relative" style={{ height: virtualizer.getTotalSize() }}>
+        {virtualizer.getVirtualItems().map(({ index, key, size, start }) => {
+          const series = data?.[index] ?? fetchNext();
```
harshithmohan, line 57:
```suggestion
          const series = data[index];
          if (!series) fetchNext();
```
The IDE was complaining about void function being assigned, also data is never undefined, it'll at least be an empty array

harshithmohan (changes requested):

src/components/Collection/Tags/TagDetailsModal.tsx
fearnlj01: It works fine 🤷 

I went the overly complex way as [the documentation](https://tanstack.com/virtual/latest/docs/api/virtualizer#getitemkey) recommends that when possible this _should_ be overridden so it's not returning just an index.

I'll remove this snippet in about an hours time as it keeps life simple, and it's not like the list of data will ever change order.

harshithmohan (approved):

PR #933, opened 2024-06-03
Misc: Add stronger overlay and colour to episode watch/hidden state

fearnlj01:


fearnlj01:
Should go from this:
image
to this:
image

Even more prominent on episodes with lighter backgrounds 👍

PR #935, opened 2024-06-11
Feat: Add edit group modal

fearnlj01:
This PR will most certainly require a proper review before merging 🙃 (my apologies)

From testing, things do at least seem to work 'okay'. My tests haven't included any nested groups at all, so I'm making no guarantees that things will work fine in this situation...

I'll also note that the SeriesTab's layout will look a little odd in terms of the amount of padding for any group with a single normal series, but additional empty series will only show a single series in the list as empty series aren't currently requested in the list of series.

It may also be worth giving consideration to either sorting the series list for the group by series ID, or forcing a ~500ms loading animation after a fetch starts for group data on the SeriesTab. As it currently stands, when changing the preferred series, the list is re-ordered and the visual jump is a little jarring.

### DIFF
--- a/src/core/react-query/group/queries.ts
+++ b/src/core/react-query/group/queries.ts
@@ -2,7 +2,7 @@ import { useInfiniteQuery, useQuery } from '@tanstack/react-query';
 
 import { axios } from '@/core/axios';
 
-import type { GroupSeriesRequestType, GroupsInfiniteRequestType } from '@/core/react-query/group/types';
+import type { GroupsInfiniteRequestType } from '@/core/react-query/group/types';
 import type { ListResultType } from '@/core/types/api';
 import type { CollectionGroupType } from '@/core/types/api/collection';
 import type { SeriesType } from '@/core/types/api/series';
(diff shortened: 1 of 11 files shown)


fearnlj01:
Just adding some screenshots, since they're far less of a pain to review at a glance...

#### Edit Group Modal [Name] - "Default" state
image

#### Edit Group Modal [Name] - "Editing" state
image

#### Edit Group Modal [Series]
image
_Perhaps it's not worth providing any special padding for the scrollbar here?_


#### Collection - Poster View
image
_It looks exactly the same as a single series group / a series when hovered over_

harshithmohan (changes requested):
Memoize both the tabs

src/core/react-query/group/queries.ts
```diff
@@ -34,9 +34,10 @@ export const useGroupsInfiniteQuery = (params: GroupsInfiniteRequestType) =>
     },
   });
 
-export const useGroupSeriesQuery = (groupId: number, enabled = true) =>
+export const useGroupSeriesQuery = ({ enabled = true, groupId, sorted = false }: GroupSeriesRequestType) =>
   useQuery<SeriesType[]>({
     queryKey: ['group-series', groupId],
     queryFn: () => axios.get(`Group/${groupId}/Series`),
+    select: data => (sorted ? data.sort((a, b) => (a.IDs.ID > b.IDs.ID ? 1 : -1)) : data),
```
harshithmohan: Why not sort them in the component itself with a useMemo?

harshithmohan (changes requested):

src/core/react-query/group/queries.ts
fearnlj01: Already resolved as requested! Had a moment and completely forgot that this would impact anything that tried using this query...

PR #936, opened 2024-06-12
Misc: Fetch all files belonging to manually linked series

fearnlj01:
I *think* this might (naïvely) fix the issue that is screenshot below, as per the `#support` channel on Discord... But I can't test it as I have nothing manually linked, nor do I properly understand how the utilities table works.

I suspect that the performance implications of my tweak may present an issue, especially if someone were to collect different resolutions for the same episodes across a series like Meitantei Conan?

image
image


revam (approved):
Seems fine to me. As long as it's not in the 10k+ range of files then the browser can probably handle it just fine.

harshithmohan (changes requested):
The UI possibly needs to change to dual pane like release management and list should be virtualized. So wait for EC

harshithmohan:
Well, I guess we can do this for now

PR #953, opened 2024-08-05
fix: update settings and series pages

revam:
Updated the settings and series pages after the TMDB PR.

The settings now have overhauled language settings, though the source order selector aren't implemented. Also, all TMDB settings (except the user api key override) are exposed.

On the series page the linked TMDB entity will indicate if it's a movie or show link, and the external link for both should work.

I couldn't get eslint/dprint to play nice on the last change, so I'm opening this as a draft.
@hidden4003 or @harshithmohan feel free to complete or edit the PR to your heart's content.


src/components/Dialogs/LanguagesModal.tsx
```diff
@@ -10,16 +10,23 @@ import { useSettingsQuery } from '@/core/react-query/settings/queries';
...
   onClose: () => void;
 };
 
 function LanguagesModal({ onClose, type }: Props) {
   const settings = useSettingsQuery().data;
   const LanguagePreference = useMemo(
-    () => (type === 'Episode'
-      ? settings.EpisodeLanguagePreference
-      : settings.LanguagePreference),
+    () => {
```
revam, line 20: these switches exist because I couldn't use nested ternaries. Feel free to keep or remove them.

src/pages/collection/series/SeriesOverview.tsx
```diff
@@ -82,7 +82,34 @@ const SeriesOverview = () => {
...
+                            key={`${site}-${subType}-${id}`}
+                          >
+                            <SeriesMetadata
+                              site={site}
+                              id={id}
+                              type={subType as 'Movie' | 'Show'}
+                              seriesId={series.IDs.ID}
+                            />
+                          </div>
+                        )),
```
revam, line 110: And this and the next line was what dprint/eslint was fighting over.

src/components/Collection/SeriesMetadata.tsx
```diff
@@ -3,58 +3,47 @@ import { mdiCloseCircleOutline, mdiOpenInNew, mdiPencilCircleOutline, mdiPlusCir
...
         deleteTvdbLink(seriesId);
         break;
       case 'TMDB':
-        switch (type) {
-          case 'Show':
-            break;
-          default:
-            break;
-        }
+        deleteTmdbLink(seriesId);
```
revam, line 46: i think you just broke it for movies. \*_looks a but further down_\* nope, but please rename the helper so it's less confusing that it works for both shows and movies. 😔

src/core/types/api/series.ts
```diff
@@ -44,10 +44,12 @@ export type SeriesIDsType = {
   TopLevelGroup: number;
   AniDB: number;
   TvDB: number[];
-  TMDB: number[];
   MAL: number[];
   TraktTv: number[];
-  AniList: number[];
+  TMDB: {
```
revam, line 49: i forgot to update this, yes.

src/pages/settings/tabs/MetadataSitesSettings.tsx
```diff
@@ -187,7 +187,7 @@ function MetadataSitesSettings() {
       <div className="border-b border-panel-border" />
 
       <div className="flex flex-col gap-y-6">
-        <div className="flex h-[2.149rem] items-center font-semibold">TvDB Options</div>
+        <div className="flex h-[2.149rem] items-center font-semibold">TVDB Options</div>
```
revam: the change from TVDB to TvDB 8 did was intentional, but if you don't want it then i won't pursue it further.

harshithmohan: The official name has all uppercase so I kept that

src/components/Collection/SeriesMetadata.tsx
harshithmohan: It would be confusing if it mentioned one or the other.
The current one just means it deletes *a* tmdb link, doesn't matter which one

src/components/Collection/SeriesMetadata.tsx
```diff
@@ -3,58 +3,47 @@ import { mdiCloseCircleOutline, mdiOpenInNew, mdiPencilCircleOutline, mdiPlusCir
...
-) => {
-  const linkId = Array.isArray(id) ? id[0] || null : id ?? null;
+type Props = {
+  id?: number;
+  seriesId: number;
+  site: 'AniDB' | 'TMDB' | 'TvDB' | 'TraktTv';
+  type?: 'Movie' | 'Show';
+};
 
+const MetadataLink = ({ id, seriesId, site, type }: Props) => {
```
revam, line 16: > was refering to the name of the hook, not the local variable. 😔

harshithmohan: You should've mentioned it properly 😔 
Fixed it

PR #983, opened 2024-08-15
Refactor topnav menu items

hidden4003:


src/components/Layout/ExternalLinkMenuItem.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Icon } from '@mdi/react';
+
+type ExternalLinkMenuItemProps = {
+  url: string;
+  name: string;
+  icon: string;
+};
+
+export const ExternalLinkMenuItem = React.memo(({
```
harshithmohan, line 10: Make the components default exports. Hot reload doesn't work otherwise.

src/components/Layout/LinkMenuItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Icon } from '@mdi/react';
+import cx from 'classnames';
+
+type LinkMenuItemProps = {
+  icon: string;
+  onClick: () => void;
+  path: string;
+  text: string;
+};
+export const LinkMenuItem = React.memo((props: LinkMenuItemProps) => {
```
harshithmohan, line 12: Make the components default exports. Hot reload doesn't work otherwise.

src/components/Layout/ExternalLinkMenuItem.tsx
hidden4003: didn't notice it did that as I used extract

PR #1003, opened 2024-08-21
Various Hover Tweaks

natyusha:
- remove hover from disabled buttons
- allow button tooltip placement to be defined with a fallback to "top"
- move theme's logo css below variables from the tailwind config
- add hover to series page tabs

### DIFF
--- a/src/components/Input/Button.tsx
+++ b/src/components/Input/Button.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import type { PlacesType } from 'react-tooltip';
 import { mdiLoading } from '@mdi/js';
 import { Icon } from '@mdi/react';
 import cx from 'classnames';
@@ -14,6 +15,7 @@ type Props = {
   loading?: boolean;
   loadingSize?: number;
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  place?: PlacesType;
   submit?: boolean;
   tooltip?: string;
 };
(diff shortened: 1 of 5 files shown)


src/components/Input/Button.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import type { PlacesType } from 'react-tooltip';
```
harshithmohan, line 2: Can this be moved to below line 7 (with a one line gap after 7)?
If linter complains, move it back 😔

src/components/Input/Button.tsx
```diff
@@ -14,6 +15,7 @@ type Props = {
   loading?: boolean;
   loadingSize?: number;
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  place?: PlacesType;
```
harshithmohan, line 18: Rename to `tooltip-place`

harshithmohan (changes requested):

src/components/Input/Button.tsx
natyusha: Unfortunately: `react-tooltip type import should occur before import of @mdi/js  import/order`

src/components/Input/Button.tsx
natyusha: Used `tooltipPlace` to avoid a hyphen in the property names, should be close enough I think

PR #1012, opened 2024-08-23
Expose minimum server requirement in builds, discord, and GH releases

revam:
- Moved the min. server requirement to the `package.json`, making it easier to find when needed to be updated (impo), in addition to easier for both vite and the GH Action runner to programmatically find and use.

- Added the min. server requirement to the `version.json` produced and included with the builds, in addition to the discord embeds and GH pre-releases.

### DIFF
--- a/.github/workflows/Release-Dev-Auto.yml
+++ b/.github/workflows/Release-Dev-Auto.yml
@@ -81,7 +81,7 @@ jobs:
             **Version**: `${{ steps.semver.outputs.version_tag }}`
             **Minimum Server Version Required**: `${{ env.MIN_SERVER_REQUIREMENT }}`
 
-            Update from the Web UI or click [here](https://github.com/${{ github.repository }}/releases/tag/v${{ steps.semver.outputs.version_tag }}) to download!
+            Update from the Web UI or download it from this release!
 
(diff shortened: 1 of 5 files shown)


.github/workflows/Release-Dev-Auto.yml
```diff
@@ -71,7 +77,14 @@ jobs:
           tag_name: v${{ steps.semver.outputs.version_tag }}
           prerelease: true
           fail_on_unmatched_files: true
-          body_path: ./changelog.txt
+          body: |
+            **Version**: `${{ steps.semver.outputs.version_tag }}`
+            **Minimum Server Version Required**: `${{ env.MIN_SERVER_REQUIREMENT }}`
+
+            Update from the Web UI or click [here](https://github.com/${{ github.repository }}/releases/tag/v${{ steps.semver.outputs.version_tag }}) to download!
```
revam, line 84: This text can probably be edited or straight out removed.

ElementalCrisis:
As discussed on Discord, we will not be implementing these changes as we do not want users being able to run older daily versions of the server and/or Web UI.

If you want to use daily, always have to be on the latest.

PR #1051, opened 2024-09-05
Misc changes (Part ??)

revam:
feel free to not merge this if you disagree with my changes. 😔

### DIFF
--- a/src/components/Collection/Tmdb/AniDBEpisode.tsx
+++ b/src/components/Collection/Tmdb/AniDBEpisode.tsx
@@ -34,7 +34,7 @@ const AniDBEpisode = React.memo(({ episode, extra, isOdd, onIconClick }: Props) 
       data-tooltip-content={episode.Name}
     >
       <div className="line-clamp-1 text-xs font-semibold opacity-65">
-        {episode.AniDB?.AirDate ?? 'Airdate Unknown'}
+        {episode.AniDB?.AirDate ?? '????-??-??'}
       </div>
       <div className="line-clamp-1">
         {episode.Name}
--- a/src/components/Collection/Tmdb/EpisodeSelect.tsx
+++ b/src/components/Collection/Tmdb/EpisodeSelect.tsx
@@ -100,7 +100,7 @@ const EpisodeSelect = React.memo((props: Props) => {
               data-tooltip-content={tmdbEpisode?.Title ?? ''}
             >
               <div className="line-clamp-1 text-xs font-semibold opacity-65">
-                {tmdbEpisode?.AiredAt ?? '????-??-??'}
+                {tmdbEpisode?.AiredAt ?? (!tmdbEpisode ? 'XXXX-XX-XX' : '????-??-??')}
               </div>
               <div className="line-clamp-1">
                 {tmdbEpisode?.Title ?? 'Not Linked'}
--- a/src/components/Collection/Tmdb/EpisodeSelect.tsx
+++ b/src/components/Collection/Tmdb/EpisodeSelect.tsx
@@ -103,7 +103,7 @@ const EpisodeSelect = React.memo((props: Props) => {
                 {tmdbEpisode?.AiredAt ?? '????-??-??'}
               </div>
               <div className="line-clamp-1">
-                {tmdbEpisode?.Title ?? 'Entry Not Linked'}
+                {tmdbEpisode?.Title ?? 'Not Linked'}
               </div>
             </div>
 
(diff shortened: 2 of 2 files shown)


src/pages/collection/series/TmdbLinking.tsx
```diff
@@ -179,7 +179,7 @@ const TmdbLinking = () => {
           TmdbShowID: tmdbId,
           TmdbEpisodeID: overrideId,
           Index: index,
-          Rating: (overrideId === 0 ? MatchRatingType.None : MatchRatingType.UserVerified),
+          Rating: MatchRatingType.UserVerified,
```
harshithmohan, line 182: Why? Why do you think it existed in the first place and why did you remove it?

src/pages/collection/series/TmdbLinking.tsx
```diff
@@ -216,6 +216,13 @@ const TmdbLinking = () => {
           [] as TmdbEpisodeXrefMappingRequestType[],
         );
 
+        // If we're not giving the server any clues about which series to link
+        // then we need to first create the auto links before sending the
+        // mappings.
+        if (isNewLink && every(newMappings, link => link.TmdbID === 0)) {
+          await createAutoLinks({ tmdbShowID: tmdbId });
+        }
+
```
harshithmohan, line 225: you could refactor the above if-else and this to have a single `createAutoLinks` line

src/components/Collection/Tmdb/AniDBEpisode.tsx
```diff
@@ -34,7 +34,7 @@ const AniDBEpisode = React.memo(({ episode, extra, isOdd, onIconClick }: Props)
       data-tooltip-content={episode.Name}
     >
       <div className="line-clamp-1 text-xs font-semibold opacity-65">
-        {episode.AniDB?.AirDate ?? 'Airdate Unknown'}
+        {episode.AniDB?.AirDate ?? '????-??-??'}
```
harshithmohan, line 37: Is that more readable then "Unknown Airdate"?

harshithmohan (changes requested):

src/components/Collection/Tmdb/AniDBEpisode.tsx
ElementalCrisis: This will confuse people.

src/pages/collection/series/TmdbLinking.tsx
revam: It is no longer needed. In fact, it's getting in the way now. Because the server now distinguished between auto-matched (AKA "none") and overridden (AKA "user verified") non-links, where the override will prevent the automagic linking from changing the link, while the auto-matched will not. This change is just to brink the UI in line with what the server will return once the match is finalized.

src/pages/collection/series/TmdbLinking.tsx
revam: I'll move this change (after fixing it) and the above to their own PR.

src/components/Collection/Tmdb/AniDBEpisode.tsx
revam: > Is that more readable than "Unknown Airdate"?

Please read the sauce code again — it says **Airdate Unknown**, not **Unknown air-date**. Impo, I found it a more coherent experience to display all "?"s in the same width and size as the formatted date, rather than using an improperly formed sentence with one made-up word ("airdate" instead of "air-date") and one proper word, written in reverse order, putting the descriptive term after the word it describes.

It was triggering my OCD _so_ **hard** that I submitted this PR. But then again, feel free to not merge this if you disagree with the change.

harshithmohan: Made up?

https://www.merriam-webster.com/dictionary/airdate

harshithmohan: You can change it to fix the order, but not '?'

hidden4003: Bunch of question marks will lead to people report that something is broken, you can't show information like that to users.

revam: > Made up?
> 
> https://www.merriam-webster.com/dictionary/airdate

Both my phone and pc doesn't recognise it as a word from it's included dictionary. I'll leave it at that.

revam: > You can change it to fix the order, but not '?'

any suggestions?

harshithmohan: Unknown airdate

harshithmohan: > > Made up?
> > https://www.merriam-webster.com/dictionary/airdate
> 
> Both my phone and pc doesn't recognise it as a word from it's included dictionary. I'll leave it at that.

That's like saying you'll change "color" to "colour" because it's not recognized. I did check it in Google before putting "Airdate" there because I had the same doubt.

ElementalCrisis:
So I take it we can close this PR since the changes are not happening?

PR #1052, opened 2024-09-06
don't force refresh themes all the time

revam:


index.html
```diff
@@ -8,7 +8,7 @@
   </head>
   <body>
     <!-- Theme must be loaded after main.css -->
-    <link rel="stylesheet" type="text/css" href="/api/v3/WebUI/Theme.css?forceRefresh=true" />
+    <link rel="stylesheet" type="text/css" href="/api/v3/WebUI/Theme.css" />
```
hidden4003, line 11: why not?

revam: So we don't put unnecessary load on the server by **always** bypassing the server-side, in-memory cache for the themes. It was either removing the cache or removing the force option **from the .css** resource. It will still force refresh and update the cache if you visit the settings page to view the themes instaooed themes.

This was discussed in the chat yesterday in case you missed it. 🫠

PR #1055, opened 2024-09-07
Further Refine Description Cleaning Regex

natyusha:
Took the regex from Shoko Relay's implementation and applied it here. This covers quite a few leftovers while maintaining the spirit of the current cleaner.

Some files are attached below which will show exactly what each regex is removing from all of the descriptions in the xml cache (after 1 pass).
[InfoLines.txt](https://github.com/user-attachments/files/16919953/InfoLines.txt) / [MiscLines.txt](https://github.com/user-attachments/files/16919954/MiscLines.txt) / [BBCode[i].txt](https://github.com/user-attachments/files/16919956/BBCode.i.txt)

Any other strange cases that have been overlooked can be easily found by scrolling through the file below:
[FinalResults.txt](https://github.com/user-attachments/files/16919957/FinalResults.txt)

Note that there are still some url related edge cases in the xml cache, but they should be covered by these creqs:
[19182093](https://anidb.net/admin/creq/19182093) / [19182097](https://anidb.net/admin/creq/19182097) / [19182764](https://anidb.net/admin/creq/19182764) / [19182767](https://anidb.net/admin/creq/19182767) / [19182768](https://anidb.net/admin/creq/19182768) / [19182771](https://anidb.net/admin/creq/19182771) / [19182772](https://anidb.net/admin/creq/19182772) / [19182773](https://anidb.net/admin/creq/19182773)

### DIFF
--- a/src/components/Collection/CleanDescription.tsx
+++ b/src/components/Collection/CleanDescription.tsx
@@ -4,11 +4,12 @@ import cx from 'classnames';
 // The question marks are there because people can't spell…
 const CleanInfoLinesRegex = /\b((Modified )?Sour?ce|Note( [1-9])?|Summ?ary):(?!$)([^\r\n]+|$)/img;
 
-const CleanMiscLinesRegex = /^^(\*|\u2014 (adapted|source:?|summary|translated|written)|- (translated)|~ (adapted|description|summary|translated)) ([^\r\n]+|$)/img;
+const CleanMiscLinesRegex =
+  /^^(\*|\u2014 (adapted|source:?|summary|translated|written)|- (translated)|~ (adapted|description|summary|translated)) ([^\r\n]+|$)/img;
 
 // This accounts for an AniDB API bug since BBCode is not supposed to be there
 const CleanBBCodeRegex = /\[i\](?!"The Sasami|"Stellar|In the distant| occurred in)(.*?)\[\/i\]/isg;
-const CleanExtraBBCodeRegex = /(\[i\]|\[\/i\])/ig; 
+const CleanExtraBBCodeRegex = /(\[i\]|\[\/i\])/ig;
 
 const CleanMultiEmptyLinesRegex = /\n{2,}/g;
 
--- a/src/components/Collection/CleanDescription.tsx
+++ b/src/components/Collection/CleanDescription.tsx
@@ -2,21 +2,21 @@ import React, { useMemo } from 'react';
 import cx from 'classnames';
 
 // The question marks are there because people can't spell…
-const RemoveSummaryRegex = /\b(Sour?ce|Note|Summ?ary):([^\r\n]+|$)/mg;
+const CleanInfoLinesRegex = /\b((Modified )?Sour?ce|Note( [1-9])?|Summ?ary):(?!$)([^\r\n]+|$)/img;
 
-const RemoveBasedOnWrittenByRegex = /^(\*|\u2014) (based on|written by) ([^\r\n]+|$)/img;
+const CleanMiscLinesRegex = /^^(\*|\u2014 (adapted|source:?|summary|translated|written)|- (translated)|~ (adapted|description|summary|translated)) ([^\r\n]+|$)/img;
 
-const RemoveBBCodeRegex = /\[i\](.*?)\[\/i\]/sg;
+// This accounts for an AniDB API bug since BBCode is not supposed to be there
+const CleanBBCodeRegex = /\[i\](?!"The Sasami|"Stellar|In the distant| occurred in)(.*?)\[\/i\]/isg;
+const CleanExtraBBCodeRegex = /(\[i\]|\[\/i\])/ig; 
 
-const MultiSpacesRegex = /\s{2,}/g;
+const CleanMultiEmptyLinesRegex = /\n{2,}/g;
 
-const CleanMiscLinesRegex = /^(--|~) /sg;
-
-const CleanMultiEmptyLinesRegex = /\n{2,}/sg;
+const CleanMultiSpacesRegex = /\s{2,}/g;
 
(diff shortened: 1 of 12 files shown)


src/components/Collection/CleanDescription.tsx
```diff
@@ -2,21 +2,23 @@ import React, { useMemo } from 'react';
...
 // The question marks are there because people can't spell…
-const RemoveSummaryRegex = /\b(Sour?ce|Note|Summ?ary):([^\r\n]+|$)/mg;
+const CleanInfoLinesRegex = /\b((Modified )?Sour?ce|Note( [1-9])?|Summ?ary):(?!$)([^\r\n]+|$)/img;
 
-const RemoveBasedOnWrittenByRegex = /^(\*|\u2014) (based on|written by) ([^\r\n]+|$)/img;
-
-const RemoveBBCodeRegex = /\[i\](.*?)\[\/i\]/sg;
+// eslint-disable-next-line operator-linebreak -- Because dprint and eslint can't agree otherwise. Feel free to fix it.
+const CleanMiscLinesRegex =
+  /^(\*|\u2014 (adapted|source:?|summary|translated|written)|- (translated)|~ (adapted|description|summary|translated)) ([^\r\n]+|$)/img;
```
revam, line 9: Which character is unicode 2014, and why is it needed here? 

Also just combine the unicode, dash, and tilde into a single 1 character block, and the other words into another, and allow the colon to be present for al- _you know what, I'll just give you a code suggestion instead of trying to explain it in plain English. Here;
```suggestion
  /^(\*|[\u2014\-~] (adapted|source|description|summary|translated|written):?) ([^\r\n]+|$)/img;
```

src/components/Collection/CleanDescription.tsx
```diff
@@ -2,21 +2,23 @@ import React, { useMemo } from 'react';
...
-const MultiSpacesRegex = /\s{2,}/g;
+// This accounts for an AniDB API bug since BBCode is not supposed to be there
+const CleanBBCodeRegex = /\[i\](?!"The Sasami|"Stellar|In the distant| occurred in)(.*?)\[\/i\]/isg;
+const CleanExtraBBCodeRegex = /(\[i\]|\[\/i\])/ig;
 
-const CleanMiscLinesRegex = /^(--|~) /sg;
+const CleanMultiEmptyLinesRegex = /\n{2,}/g;
 
-const CleanMultiEmptyLinesRegex = /\n{2,}/sg;
+const CleanMultiSpacesRegex = /\s{2,}/g;
```
revam, line 17: Should probably have the s flag, to not collapse newlines unintentionally, even if this is ran after the regex to reduce multiple new lines into a single new line.

revam (changes requested):
The rest except these two changes/comments seemed fine. Haven't tested any of this on real data, so I'll just trust @natyusha has already done that extensively enough already.

hidden4003:
Me and @ElementalCrisis have already decided on this, it can be added only if there is an option to turn it off. Can't merge until option is there.

src/components/Collection/CleanDescription.tsx
natyusha: hmm, didn't think the s flag did anything there (added it anyways)

src/components/Collection/CleanDescription.tsx
natyusha: u2014 is an emdash (—) which accounts for ~300 descriptions which have written by entries (my example file above only listed the first match for each regexp per description).

The suggested changes do make it slightly less clear which keywords apply to which special character but I do prefer it like this thanks.

natyusha:
I see no reason to delay merging this when the automatic description cleaning has already been in effect for over a year. All this does is address some areas that were lacking a little coverage and remove some errant BBCode. Unless the description cleaning is removed entirely keeping this PR sitting in limbo makes no logical sense. I find myself envisioning a scenario where someone just woke up from a year long hibernation. They then proceeded to check the descriptions in the WebUI for the first time since then with a shocked look on their face.

Also, there is nothing wrong with adding an option and I actually have both the "Info" and "Misc" lines setup to be optional in Shoko Relay.

hidden4003:
I can only agree on the regexes that remove excessive white space, excessive new lines or the line at the end that anidb has with a user name who created the description.

Rest of them regarding based on/written by are highly subjective and to me personally are not something that I think should be removed from the description.

Which is why there has to be a setting.

natyusha:
The way it works now there are a huge amount of descriptions that only have that "* Based on" section (or something similar) and by allowing the regex to remove it you invoke the fallback to the TMDB description.

In fact, it was two weeks ago that this was discussed in Discord. During which I was told that the "* Based on" section was fine to remove. Who do I believe when receiving mixed messages like this?

ElementalCrisis:
I think we should keep the minimal cleaning we had previously but make this more advanced cleaning the option. 

Basically something like "Clean AniDB Descriptions" with some info pertaining to what would be cleaned. 

@natyusha I can add the setting this week if you'd like me to.

natyusha:
Go for it, you're the one who said it was fine to remove the "* based on" lines in the first place 😄

natyusha:
Writing down what is considered "minimal" cleaning and what isn't would be great. From my understanding it is just the lines that start with a "*" that people might want right?

natyusha:
You guys can deal with it however you see fit 😄 

I highly suggest fixing the broken BBCode remnants though.

src/components/Collection/CleanDescription.tsx
natyusha: I just thought it only affected the `.` wildcard no idea if that is true here vs what I'm used to using though

hidden4003: With single line or DOTALL flag enabled `.` will match newline characters but you are not using `.` in these expressions so the flag does not have an effect.

revam: then I'm misremembering.

PR #1057, opened 2024-09-08
don't show tvdb ids linked through tmdb data

revam:


src/pages/collection/series/SeriesOverview.tsx
```diff
@@ -31,9 +31,17 @@ const SeriesOverview = () => {
   const { seriesId } = useParams();
 
   const seriesQuery = useSeriesQuery(toNumber(seriesId!), { includeDataFrom: ['AniDB', 'TMDB'] }, !!seriesId);
-  const series = useMemo(() => seriesQuery?.data ?? {} as SeriesType, [seriesQuery.data]);
+  const series = useMemo(
+    () =>
+      seriesQuery?.data
+        ?? {
```
revam, line 37: In case someone reading this is wondering, this will prevent any errors where we try to access an undefined property as if it were defined and make the runtime throw as a result.

PR #1061, opened 2024-09-09
fix sorting of the manually linked files

revam:
Fix sorting of the manually linked files by episode type, episode number, followed by file creation date.

If the manually linked components were refactored to use the episode endpoint first and foremost then we might have avoided to do this client-side sorting. But. I'm not going to put in the effort to refactor them all at this point in time, so this band-aid will do for now. At least until somebody puts in the time and effort to do the work.


harshithmohan:
The UI for this page and will change to the dual pane one like release management and then the episode list will be paginated, and then we cannot sort it. What we need to do is to get the episodes instead of files. Is that possible? 🤔

harshithmohan:
Not needed due to #1067

PR #1062, opened 2024-09-09
Add  '🖋 Upload Theme' button to settings

revam:
`feat`: Add upload theme button to settings.

Added a new '🖋 Upload Theme' button to the 'Theme Options' section of the 'General' Settings tab in the settings page. With it you can upload JSON-only themes, CSS-only themes, and JSON-based online themes, all without having access to the file system of the server. The themes are also auto-applied in preview mode after they're successfully uploaded, so you can preview the theme right away before saving.

**P.S.**: _If somebody have issues with the implementation, then be my guest changing it, as long as user experience remains the same in the end._

`misc`: Bump min. server version.

### DIFF
--- a/src/pages/settings/tabs/GeneralSettings.tsx
+++ b/src/pages/settings/tabs/GeneralSettings.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import React, { useMemo } from 'react';
-import { mdiOpenInNew, mdiRefresh } from '@mdi/js';
+import React, { useMemo, useRef } from 'react';
+import { toast } from 'react-toastify';
+import { mdiBrushOutline, mdiOpenInNew, mdiRefresh } from '@mdi/js';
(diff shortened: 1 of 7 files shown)


src/pages/settings/tabs/GeneralSettings.tsx
```diff
@@ -27,8 +32,49 @@ function GeneralSettings() {
...
+    uploadTheme({ file }, {
+      async onSuccess(data) {
+        await themesQuery.refetch();
+
+        let path = themePathHref.value;
+        const index = path.indexOf('?');
+        if (index !== -1) path = path.substring(0, index);
+        themeUpdateCounter += 1;
+        path += `?updateCount=${themeUpdateCounter}`;
+        themePathHref.value = path;
```
harshithmohan, line 71: do this with `new URL(themePathHref).value` and corresponding properties, so that it's more readable

harshithmohan (changes requested):

PR #1066, opened 2024-09-10
Further Refine Description Cleaning Regex Redux

natyusha:
Applied changes from: https://github.com/ShokoAnime/Shoko-WebUI/pull/1055 to the new "Filter AniDB descriptions" option (thanks avael).

Also, added "whitespace-pre-line" to the description text so it won't just be a large block even when line breaks are present in the description.

### DIFF
--- a/src/components/Collection/CleanDescription.tsx
+++ b/src/components/Collection/CleanDescription.tsx
@@ -4,23 +4,23 @@ import cx from 'classnames';
 import { useSettingsQuery } from '@/core/react-query/settings/queries';
 
 // The question marks are there because people can't spell…
-const RemoveSummaryRegex = /\b(Sour?ce|Note|Summ?ary):([^\r\n]+|$)/mg;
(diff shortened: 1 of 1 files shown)


hidden4003 (approved):

src/components/Collection/CleanDescription.tsx
```diff
@@ -4,23 +4,23 @@ import cx from 'classnames';
 import { useSettingsQuery } from '@/core/react-query/settings/queries';
 
 // The question marks are there because people can't spell…
-const RemoveSummaryRegex = /\b(Sour?ce|Note|Summ?ary):([^\r\n]+|$)/mg;
+const CleanInfoLinesRegex = /\b((Modified )?Sour?ce|Note( [1-9])?|Summ?ary):(?!$)([^\r\n]+|$)/img;
```
revam, line 7: only question i have is why is the negative lookahead for the end of text needed?

revam (commented):
the rest of the regexes look fine, impo.

src/components/Collection/CleanDescription.tsx
natyusha: For something like this: https://anidb.net/anime/9283

PR #1136, opened 2024-11-06
Modify Metadata Linking Header / AvDump Series Select QoL / Group Page Edit Button

natyusha:
Tried to make it a little bit more clear what the new number means and put a couple other proposals in the ss too (i actually prefer alt 2):
firefox_2024-11-05_23-19-55


harshithmohan:
We can do alt 2 but change the colours

natyusha:
Here is the condensed version:
firefox_2024-11-06_04-58-32

natyusha:
Edit Group Button:
firefox_2024-11-09_00-07-12

PR #1141, opened 2024-11-15
Keep filterId in links to remember filter

Jonas-lauwers:


harshithmohan (commented):
The series page URLs should never have filter or group IDs because nothing is being filtered. This is just wrong.
If you want to be able to go back like that, open the series in a new tab.

And I blame @revam for asking to create a PR before properly understanding what is going to be changed or if it should be changed

PR #1147, opened 2024-11-24
Feat: Add buttons to copy the ED2K hash to the clipboard in the link format

fearnlj01:
Implements a clickable clipboard icon next to wherever the ED2K hash is presented, allowing for the hash to be copied in the standard ED2K link format.


fearnlj01:
Screenshots of the changes are provided below:

File Info (Although it might want removing from here for the sake of consistency?):
image

File Search:
image

PR #1148, opened 2024-11-26
Feat: Add `DownloadMissingTmdbPeople` as available action

fearnlj01:
Makes the recently added server endpoint more readily available.

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -29,7 +29,7 @@ export function isDebug() {
   return DEV;
 }
 
-export const minimumSupportedServerVersion = '4.2.2.182';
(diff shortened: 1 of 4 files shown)


src/core/util.ts
```diff
@@ -29,7 +29,7 @@ export function isDebug() {
   return DEV;
 }
 
-export const minimumSupportedServerVersion = '4.2.2.182';
+export const minimumSupportedServerVersion = '5.0.0.61';
```
harshithmohan, line 32: No. Figure out a way to use `useIsFeatureSupported`

harshithmohan (changes requested):

PR #1162, opened 2024-12-30
fix: de-weebify actors

revam:


da3dsoul:
Why

revam:
I can revert it for the UI part, but the server side role value changed, so the web UI needed to be updated accordingly.

And if you're wondering why I did three forced pushes, then it's because initially I was going to go full no backwards compatibility with stable, then I later changed my mind to try and be compatible with both stable and daily, but then I remembered @harshithmohan [had said](https://discord.com/channels/96234011612958720/472272352231686144/1315394311726043249) the web UI was currently in a state where it was fine to break compatibility with stable, so I tried to reverse the changes to make it compatible with stable, failed miserably, and had to fix the changes. And by the last commit the code was the same as the initial commit again.

da3dsoul:
No, I was wondering why you renamed it in general

revam:
2-3am coding session train-of-thought thought it was better to call the actors as `Actor(s)` (proper English word) instead of `Seiyuu(s)` (weeb-ish engrish word).

Renari:
FateKiritsuguGIF

PR #1171, opened 2025-01-20
Add custom tag modal

revam:
I made sure the functionality of it is working as intended according to how I would assume it should work before I started this PR. As for all other aspects, feel free to roast it.

The final UI doesn't quite match the mock-ups, since I discovered while building/testing that a) the mock-up is old and the modal/dialog style is outdated compared to the current code base, and b) the lack of any indicator showing which tags are currently added/in-use, which I addressed, and c) the lack of any toggle for adding/removing tags from use on a series, which I also addressed.

### DIFF
--- a/src/components/Collection/SeriesTopPanel.tsx
+++ b/src/components/Collection/SeriesTopPanel.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useMemo, useState } from 'react';
 import { useParams } from 'react-router';
+import { mdiTagPlusOutline } from '@mdi/js';
+import Icon from '@mdi/react';
 import { toNumber } from 'lodash';
 
 import BackgroundImagePlaceholderDiv from '@/components/BackgroundImagePlaceholderDiv';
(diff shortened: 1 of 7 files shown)


harshithmohan (commented):
The modal doesn't need to be stored in redux store if it's only being used from one place.

src/components/Collection/SeriesTopPanel.tsx
```diff
@@ -78,7 +94,15 @@ const SeriesTopPanel = React.memo(({ series }: { series: SeriesType }) => {
...
           transparent
+          options={
+            <div className="flex gap-x-2">
+              <Button onClick={handleEditTagsClickHandler} tooltip="Edit Tags">
+                <Icon className="text-panel-icon-important" path={mdiTagPlusOutline} size={1} />
+              </Button>
+            </div>
+          }
         >
+          <CustomTagModal seriesId={Number(seriesId)} show={showTagModal} onClose={handleCloseModal} />
```
harshithmohan, line 105: use `toNumber` for consistency

harshithmohan (changes requested):
Other than the comments. The flow is confusing.

Separate the add/edit functionality into a separate modal and keep this one for just selecting the tag. Maybe we should put the add/edit away in the settings or something. But right now it's just confusing. ~~And it also needs to allow removing or adding multiple tags at once.~~ It does do that.. I was expecting I just select things and click on "save"

Tagging @ElementalCrisis for thoughts
image

revam:
It's been a week. I'm not waiting for a another technical review. Any UI/UX changes should probably be forwarded to the potential v3 design, for now I'm merging this so it's possible to create/delete/add/remove custom tags _at all_.

PR #1179, opened 2025-02-04
QoL: Add "Rename/Move" action to "Edit Group" and "Edit Series" modal

pokyunn:
This allow to call default renamer, on related files from Collection tab, making easier to rename files after organized in shoko groups
image


ElementalCrisis (approved):

harshithmohan:
@revam maybe add a dedicated endpoint for relocating files using series ID?
Because getting the file IDs and then sending them back to the server will cause issues for bigger groups/series.

pokyunn:
@harshithmohan I think, this would be a good addition, but I don't see the problem because  (at least what i tested), it works the same wat the utility-renamer screen works 
- Fetch series episodes
- Send them back to relocate
image

harshithmohan:
The issue occurs when there are >1000 file IDs, then they need to be divided. But my point is, why get full data for 1500 episodes just for their file IDs when we can just ask the server to do the work for us.

harshithmohan:
or we just keep nagging @revam to add it

revam:
i can add it, but not rn.

pokyunn:
@revam i created the endpoint based on /Series/{seriesID}/File/Rescan when you have time check it
https://github.com/ShokoAnime/ShokoServer/pull/1224

@harshithmohan updated to use the endpoint

harshithmohan:
https://github.com/ShokoAnime/Shoko-WebUI/blob/master/src/hooks/useIsFeatureSupported.ts

Now use this and disable the button/show a tooltip when server version isn't supported :pensive:

pokyunn:
Message added.

src/components/Collection/Series/EditSeriesTabs/FileActionsTab.tsx
```diff
@@ -10,6 +16,10 @@ type Props = {
 const FileActionsTab = ({ seriesId }: Props) => {
   const { mutate: rehashSeriesFiles } = useRehashSeriesFilesMutation(seriesId);
   const { mutate: rescanSeriesFiles } = useRescanSeriesFilesMutation(seriesId);
+  const { mutate: relocateSeriesFiles } = useRelocateSeriesFilesMutation(seriesId);
+  const unsuported = () => {
```
harshithmohan, line 20: typo, but also maybe name it better
```suggestion
  const showUnsupportedToast = () => {
```

src/core/react-query/group/mutations.ts
```diff
@@ -57,3 +58,19 @@ export const useMoveGroupMutation = () =>
...
+export const useRelocateGroupFilesMutation = (groupId: number) =>
+  useMutation({
+    mutationFn: async () => {
+      const targetSeries = await axios.get<unknown, SeriesType[]>(`Group/${groupId}/Series`, {
+        params: { recursive: true },
+      });
+
+      return Promise.all(
+        targetSeries.map(
+          serie => axios.post(`Series/${serie.IDs.ID}/File/Relocate`),
```
harshithmohan, line 71:
```suggestion
          series => axios.post(`Series/${series.IDs.ID}/File/Relocate`),
```

harshithmohan (changes requested):
Only 2 changes, everything else looks fine

PR #1189, opened 2025-03-07
fix: show both episodes & files, which may differ

revam:
Before:
image

After:
image


ElementalCrisis:
Looks good! 

My only suggestion would be to hide files when episode and file count match.

PR #1193, opened 2025-03-26
Add id to series background image div.

ElementalCrisis:
Adds an id tag to the div containing the series background image so users can manipulate the div in a custom theme.


revam (commented):
idea: add unique ids (e.g. `series-page`, `collection-page`, `dashboard-page`  etc.) and maybe a class (e.g. `page`) used on the same element as the page id for every page template, optionally with a `data-item-id` data tag for e.g. series pages, etc.

ElementalCrisis:
Yeah its not a bad idea, something we can expand on in the future.

Going to merge this but will make a note of this in the future.

PR #1215, opened 2025-08-01
feat: add simple anidb anime redirect

revam:
I'm tired of searching for the anime in my collection by name when I already know the AniDB anime id. :pensive:

I couldn't care less of how the redirect page looks, so I just mashed together something based on the login page. It is based on the login page and not main page since I didn't want the main page frame on the redirect page, but if somebody has a burning passion that says it **N E E D S** to have the main page frame then feel free to edit this PR to change it. :slightly_smiling_face:

### DIFF
--- a/src/core/react-query/series/queries.ts
+++ b/src/core/react-query/series/queries.ts
@@ -39,6 +39,13 @@ export const useSeriesQuery = (
     enabled,
   });
 
+export const useSeriesByAnidbIdQuery = (anidbId: number, enabled = true) =>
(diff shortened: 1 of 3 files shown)


harshithmohan:
https://github.com/user-attachments/assets/86736c68-31ca-4be0-bb1f-8aa3b86bb654

uhhh.. fix that?
The error always shows, the spinner is missing the colour and also in wrong size.
Can we redirect the user to the main page instead of the error page and just show a toast?

PR #1288, opened 2026-03-01
Add provider-based linking UI for unrecognized files (part 1)

revam:
Added the initial PoC of the provider-based linking UI alongside the existing manual linking UI. The current PoC is void of provider configuration support, editing the provider found release info. Both arriving in a latter commit. It is currently marked as a β in the UI until it has been further fleshed out and tested, but what is there is currently working as intended during initial testing.


src/components/Dialogs/ProviderInfoModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+type Props = {
+  show: boolean;
+  provider: HashProviderInfoType | ReleaseProviderInfoType | null;
+  onClose: () => void;
+};
+
+const ProviderInfoModal = (props: Props) => {
+  const { onClose, provider, show } = props;
+
+  const onKeyboard = useEventCallback((event: KeyboardEvent) => {
```
hidden4003: Since we have react compiler I believe this can be just a normal function, it also has no dependencies so I don't see what `useEventCallback` achieves here, I would prefer to have absolute minimum of this `useEventCallaback` in the code as possible as it is an old hack from 2018.

hidden4003 (commented):
[LinkFilesWithProvidersTab.tsx](https://github.com/ShokoAnime/Shoko-WebUI/pull/1288/changes#diff-48e939a39506df487aa106a0d253578cb36b68bebab23354e41ce9c9cc0c8f93) definitely needs some code extracted as it is way too big to be maintainable.

src/components/Utilities/Unrecognized/EditReleaseInfoModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+  useEffect(() => {
+    setReleases(cloneDeep(initialReleases));
+  }, [initialReleases]);
+
+  return (
+    <ModalPanel
+      show={releaseOrder.length > 0 && show}
+      size="sm"
+      onRequestClose={onClose}
+      shouldCloseOnEsc={false}
```
harshithmohan: You don't need to send `false` because it's false by default

src/components/Input/Checkbox.tsx
```diff
@@ -27,47 +47,47 @@ const Checkbox = memo((props: Props) => {
...
     >
       <input
         id={id}
         type="checkbox"
-        checked={!disabled && isChecked}
-        onChange={disabled ? undefined : onChange}
+        checked={isChecked}
+        disabled={disabled}
+        readOnly={readOnly}
+        onChange={disabled || readOnly ? undefined : onChange}
```
harshithmohan: Ackshually, you can just pass the onChange and onClick handlers either way without the checkbox. Disabled checkboxes won't run them

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesWithProvidersTab.tsx
```diff
@@ -0,0 +1,867 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable no-continue */
+/* eslint-disable @stylistic/function-paren-newline -- ESLint and DPrint are fighting about the formatting here. */
+/* eslint-disable @stylistic/comma-dangle -- ESLint and DPrint are fighting about the formatting here. */
```
harshithmohan: Have you tried with the latest code? The rules have changed

src/components/TabPills.tsx
```diff
@@ -0,0 +1,45 @@
...
+          className={cx(
+            'whitespace-nowrap group relative min-w-0 grow overflow-hidden shrink-0 px-4 py-4 text-center text-sm font-medium focus:z-10',
+            tab.current
+              ? '!bg-panel-toggle-background text-panel-toggle-text'
+              : 'bg-panel-background hover:bg-panel-toggle-background-hover text-panel-toggle-text-alt',
+            tabIdx === 0 ? 'rounded-l-lg' : '',
+            tabIdx === tabs.length - 1 ? 'rounded-r-lg' : '',
+          )}
+        >
+          <span>{tab.name}</span>
```
harshithmohan: What does the `<span>` do here?

src/core/react-query/release-info/mutations.ts
```diff
@@ -0,0 +1,47 @@
...
+    mutationKey: ['release-info', 'providers', providerGuid],
+    mutationFn: (provider: UpdateOneReleaseInfoProviderType) =>
+      axios.put(`/ReleaseInfo/Provider/${providerGuid}`, provider),
+    onSuccess: () => invalidateQueries(['release-info', 'providers']),
+  });
+
+export const useSubmitReleaseInfoForFileByIdMutation = () =>
+  useMutation({
+    mutationFn: ({ fileId, release }: { fileId: number, release: ReleaseInfoType }) =>
+      axios.post<unknown, void>(`/ReleaseInfo/File/${fileId}`, release),
```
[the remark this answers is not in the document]
revam: it sets the return type for the promise to `void`. the `unknown` is discorded in this case. totally cosmetic.

src/core/react-query/episode/queries.ts
```diff
@@ -26,3 +28,39 @@ export const useEpisodeAniDBQuery = (episodeId: number, enabled = true) =>
     queryFn: () => axios.get(`Episode/${episodeId}/AniDB`),
     enabled,
   });
+
+const episodePlaceholderKey = ['episode', 'anidb', 'bulk', 'placeholder'] as const;
+
+export const useEpisodeAniDbBulkQuery = (anidbIds: number[], enabled = true) => {
+  useEffect(() => () => {
```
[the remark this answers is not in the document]
revam: It shouldn't be a memo since we want an unmount effect. and double arrows signifies it's an unmount effect.

src/components/Utilities/Unrecognized/AutoSearchReleaseModal.tsx
```diff
@@ -0,0 +1,185 @@
...
+  useEffect(() => {
+    setProviders(initialProviders);
+  }, [initialProviders]);
+
+  return (
+    <ModalPanel
+      show={providers.length > 0 && show}
+      size="sm"
+      onRequestClose={onClose}
+      shouldCloseOnEsc={false}
```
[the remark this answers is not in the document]
revam: [You sure about that?](https://reactcommunity.org/react-modal/#:~:text=shouldCloseOnEsc)

src/core/types/api/file.ts
```diff
@@ -35,24 +35,24 @@ export type FileType = {
...
-  Hashes?: FileHashDigestType[];
-  MediaInfo?: ReleaseMediaInfoType;
+  Group: ReleaseGroupType | null;
+  Hashes: FileHashDigestType[] | null;
+  MediaInfo: ReleaseMediaInfoType | null;
   CrossReferences: ReleaseCrossReferenceType[];
-  Metadata?: string;
-  Released?: string;
+  Metadata: string | null;
+  Released: string | null;
```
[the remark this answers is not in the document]
revam: Because you wanted to use `field?: type` in the web UI when I did my last PR to patch up most of what was broken, and said PR didn't actually use most of the nullable fields, so I reformatted it then. I didn't think much ahead at the time as the code using said nullable fields were not checked out when I did the reformat. But then, in this PR, we're actually initializing and using them in the new provider-based linking UI, and since the server expects them to be in the body when you send it, be it null or otherwise, then it has been "reverted" back to what it initially was written as.

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesWithProvidersTab.tsx
```diff
@@ -0,0 +1,867 @@
+/* eslint-disable no-param-reassign */
```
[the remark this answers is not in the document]
revam: Object being modified is not used by react at the time it's being modified. And it's being modified in a called function, hench why the ESLint rule is complaining.

src/components/Dialogs/ConfirmationPromptModal.tsx
```diff
@@ -2,59 +2,71 @@ import React, { useState } from 'react';
...
 type Props = {
-  content: React.ReactNode;
+  children: React.ReactNode;
   onConfirm: () => void | Promise<void>;
   onClose: () => void;
   show: boolean;
   title: string;
-  confirmText?: string;
+  cancel?: React.ReactNode;
+  confirm?: React.ReactNode;
```
harshithmohan: `cancelText` and `confirmText`

harshithmohan (changes requested):

harshithmohan (changes requested):

src/components/Utilities/Unrecognized/AutoSearchReleaseModal.tsx
```diff
@@ -23,24 +23,24 @@ type AutoSearchReleaseModalProps = {
 
 const AutoSearchReleaseModal = (props: AutoSearchReleaseModalProps) => {
   const { onClose, onUpdateProviders, providers: initialProviders, show } = props;
-  const [info, setInfo] = useState<{ show: boolean, provider: ReleaseProviderInfoType | null }>(
-    () => ({ show: false, provider: null }),
+  const [info, setInfo] = useState<{ show: boolean, provider?: ReleaseProviderInfoType | undefined }>(
+    () => ({ show: false }),
```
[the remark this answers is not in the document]
revam: where is this recommendation from? since last time i read about it, it was better to do a shared state for all variables which needs to be updated in sync with each other. in this case the show and provider. things may have changed since then, but that is what I've learnt have been best to reduce rerenders back in the day.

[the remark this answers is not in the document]
harshithmohan: React (18 onwards) batches state updates which occur together.
Using objects for state only causes issues with updating them and was the reason React moved away from those when they moved to functional components. It took time to actually fix the functionality but it's fixed since 18

src/components/Utilities/Unrecognized/AutoSearchReleaseModal.tsx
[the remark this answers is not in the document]
revam: According to the source code (**not the @types package**) then the values in that example **are** the defaults. I just didn't want to link to the source code on mobile 20h ago, but if it weren't interfering with binding to the escape key then I wouldn't even have to add it.

https://github.com/reactjs/react-modal/blob/master/src/components/Modal.js#L97

src/core/types/api/file.ts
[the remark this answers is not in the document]
revam: Using providers is _optional_. Users can edit the release info without asking any providers.

(Although the edit menu isn't part of this PR, as stated in the opening post for this PR.)

src/components/Utilities/Unrecognized/UnrecognizedVideo.tsx
```diff
@@ -273,4 +179,13 @@ const LinkFilesTabVideo = (props: LinkFilesTabVideoProps): React.JSX.Element =>
...
+export default UnrecognizedVideo;
+
+function parseReleaseSource(releaseSource: ReleaseSource): string {
+  switch (releaseSource) {
+    case ReleaseSource.BluRay:
+      return 'Blu-Ray';
+    default:
+      return releaseSource;
+  }
+}
```
harshithmohan: Move this to be above the component

src/pages/settings/tabs/HashingAndReleaseSettings.tsx
harshithmohan: Why is this an IIFE??

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesWithProvidersTab.tsx
```diff
@@ -0,0 +1,853 @@
...
+              editedData.Version = 1;
+            }
+            if (editedData.FileSize == null && link.release.FileSize != null) {
+              edited = true;
+              editedData.FileSize = link.release.FileSize;
+            }
+            if (
+              editedData.OriginalFilename == null && link.release.OriginalFilename != null
+              && link.release.OriginalFilename !== ''
+            ) {
```
harshithmohan:
```suggestion
            if (editedData.OriginalFilename == null && !!link.release.OriginalFilename) {
```

harshithmohan (changes requested):

src/pages/utilities/UnrecognizedUtilityTabs/LinkFilesWithProvidersTab.tsx
hidden4003: I thought null checks need === and !=== isn't == null equivalent to just `if (xx)` ?

harshithmohan: > I thought null checks need === and !=== isn't == null equivalent to just `if (xx)` ?

`== null` checks for both undefined and null and it is allowed by the linter

`x == null` would be equivalent to `if (!!x)` but then it would go inside the if for an empty string as well which I assumed was not the intended behaviour

hidden4003 (approved):

PR #1319, opened 2026-05-13
Add 'Select All' functionality to UnrecognizedTab menu

OrbitMPGH:
I got irritated that I had to click each row individually when I had just downloaded a season with 24 episodes and wanted to AVDump it. This pull request adds a button in the unrecognized tab that simply selects all rows. The button is always visible, similar to the refresh button.

I usually don't work with front-end stuff, so feel free to send any changes you need, and I'll fix them.
I ran all the lints before pushing.

### DIFF
--- a/src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
+++ b/src/pages/utilities/UnrecognizedUtilityTabs/UnrecognizedTab.tsx
@@ -10,6 +10,7 @@ import {
   mdiEyeOffOutline,
   mdiFileDocumentEditOutline,
   mdiLinkVariantPlus,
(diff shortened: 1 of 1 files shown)


natyusha:
It is quite fast to select things in that view if you utilize the `Shift+Click` functionality.

For example: you could click the top episode -> hit `End` to scroll to the bottom -> Shift+Click the last entry

OrbitMPGH:
Haha yeah, that works perfectly. I don't really know why I didn't think to try that.

harshithmohan:
It wouldn't have worked btw. All the files are not loaded until you scroll down. They are paginated.
So selectAll would only select the files which are loaded. That is the reason why it doesn't exist.

PR #1327, opened 2026-05-22
feat: add AniDB rules confirmation to avdump file linking

### LINKED ISSUE #1262: Add Mandatory Rules Confirmation Popup for adding files to AniDB
**Problem**

Users are not following AniDB rules on what types of files can be added and are adding remuxes or their re-encodes which are considered private files and are not allowed by AniDB rules. This creates unnecessary burden for AniDB moderators as each such file entry needs to be dealt with manually.

**Proposed Solution**

Introduce an additional mandatory confirmation popup that appears before we redirect to AniDB.

**Popup Behavior & Requirements**

The popup must clearly reiterate the file upload rules from AniDB [wiki](https://wiki.anidb.net/Content:Files) and warn user that adding files that are breaking the rules will get their account banned.

A confirm/proceed button is displayed but:

- Initially disabled
- Becomes enabled only after a 10 seconds delay _and_ the user have inputted the answer to the randomly generated word riddle based on the content of the popup.
- This delay is intentional to prevent users from immediately clicking through without reading.

The user **cannot proceed** adding the files unless they wait for the button to become enabled and explicitly click Confirm.

A checkbox to delay the next popup for `<a currently unspecified amount of time>` will be displayed on the page so the user won't have to see the popup every time.

krbrs:
## Summary

Adds a mandatory AniDB rules confirmation step before opening the mass-add page from the AVDump add-files flow.

## Changes

- shows an AniDB rules warning before opening the AniDB mass-add page
- requires a 10-second wait before the user can proceed
- requires the user to answer a verification prompt before the AniDB link becomes available
- opens AniDB only from an explicit user click instead of relying on automatic popup behavior
- adds a `Don't show this warning again for 30 days` option backed by `localStorage`

## Scope

This change is limited to the AVDump add-files flow and does not modify other AniDB links in the UI.

Closes #1262

### DIFF
--- a/src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
+++ b/src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
@@ -1,10 +1,12 @@
-import React, { useEffect, useLayoutEffect, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useMemo, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
 import { mdiInformationOutline, mdiLoading, mdiMagnify, mdiOpenInNew } from '@mdi/js';
 import { Icon } from '@mdi/react';
 import { countBy, some, toNumber } from 'lodash';
 import { useDebounceValue } from 'usehooks-ts';
 
 import Button from '@/components/Input/Button';
+import Checkbox from '@/components/Input/Checkbox';
 import Input from '@/components/Input/Input';
 import ModalPanel from '@/components/Panels/ModalPanel';
 import toast from '@/components/Toast';
@@ -13,6 +15,7 @@ import { useSeriesAniDBSearchQuery } from '@/core/react-query/series/queries';
 import { useSelector } from '@/core/store';
 import { copyToClipboard } from '@/core/util';
 import { detectShow, findMostCommonShowName } from '@/core/utilities/auto-match-logic';
+import useToggleModalKeybinds from '@/hooks/useToggleModalKeybinds';
 
 type Props = {
   show: boolean;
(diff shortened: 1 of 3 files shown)


src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -52,12 +79,147 @@ const StepDescription = ({ children }: { children: React.ReactNode }) => (
...
+  }, [challenge, show]);
+
+  const isAnswerCorrect = answer.trim().toLowerCase() === challenge.answer;
+  const canProceed = secondsRemaining === 0 && isAnswerCorrect;
+
+  const handleProceed = () => {
+    if (!canProceed) return;
+
+    if (dontShowAgain) {
+      globalThis.localStorage.setItem(ANIDB_RULES_SNOOZE_KEY, `${Date.now() + ANIDB_RULES_SNOOZE_MS}`);
```
hidden4003: We use dayjs for all date related things, so just to be consistent and might also solve possible incompatibilities between browsers.

krbrs: 👌 snooze expiry was switched to dayjs

src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -52,12 +80,150 @@ const StepDescription = ({ children }: { children: React.ReactNode }) => (
   </div>
 );
 
+const AniDbRulesModal = ({
```
harshithmohan: Move this component into it's own file

src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -1,10 +1,13 @@
-import React, { useEffect, useLayoutEffect, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useMemo, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
 import { mdiInformationOutline, mdiLoading, mdiMagnify, mdiOpenInNew } from '@mdi/js';
 import { Icon } from '@mdi/react';
+import dayjs from 'dayjs';
```
harshithmohan: Import `dayjs` from `'@/core/util`
EDIT: just realized I almost wrote the whole code line anyway 😂

src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -198,7 +402,8 @@ const AvDumpSeriesSelectModal = ({ fileIds, links, onClose, show }: Props) => {
                           href={`https://anidb.net/anime/${result.ID}/release/add`}
                           target="_blank"
                           rel="noopener noreferrer"
-                          onClick={() => setClickedLink(true)}
+                          onClick={event =>
```
harshithmohan: It's an anchor tag, people can still right click to open the link and bypass the warning.
Maybe change it to a div as the link opening is handled by the onClick function anyway.

src/components/Utilities/Unrecognized/AvDumpSeriesSelectModal.tsx
```diff
@@ -52,12 +80,150 @@ const StepDescription = ({ children }: { children: React.ReactNode }) => (
...
+          <Input
+            id="anidb-rules-answer"
+            value={answer}
+            type="text"
+            placeholder="Enter your answer"
+            onChange={event => setAnswer(event.target.value)}
+            className="w-full"
+          />
+          <div className="text-sm opacity-65">
+            Wait &nbsp;
```
harshithmohan: Those are 2 spaces after "Wait"

harshithmohan (changes requested):

PR #1329, opened 2026-05-23
feat: implement plugin management UI (#1329) [no ci]

krbrs:
## Summary

Implements the initial Plugin Management feature for the Shoko WebUI based on the new plugin repository and package APIs exposed by Shoko Server.

This PR adds repository management, plugin catalog browsing, installation flows, installed plugin management, and update handling. The implementation was built from the server-side plugin package contract and Spec Kit planning artifacts generated during feature design. 

### Repository Management

* Add/remove plugin repositories
* Manual repository synchronization
* Repository listing with stale/error handling
* Duplicate repository prevention
* Collapsible repository form UX

### Plugin Catalog

* Browse available plugin packages from configured repositories
* Group releases and archive variants by package
* Search/filter support
* Installed-state awareness
* Compatibility handling for runtime/ABI mismatches
* Thumbnail support via server thumbnail endpoint
* Expandable older release history

### Plugin Installation & Updates

* Install compatible plugin releases
* Update installed plugins
* Release/archive selection support
* Installation progress and failure handling
* Restart-required state indication

### Installed Plugin Management

* List installed plugins and versions
* Enable/disable plugin versions
* Uninstall installed versions
* Distinguish built-in vs installed plugins
* Local plugin state handling


harshithmohan:
## Review Summary

This PR introduces two major features: a Plugin Management UI in settings, and an AniDB rules confirmation modal for the unrecognized files utility. The code is generally well-structured and follows existing patterns, but there are a few issues that should be addressed.

---

### 🔴 Critical

#### 1. `pnpm-workspace.yaml` contains unresolved placeholder values

**File:** `pnpm-workspace.yaml`

`allowBuilds` expects boolean values (`true`/`false`). The literal strings `"set this to true or false"` are placeholder text that will cause pnpm to reject or ignore these entries, breaking builds for anyone using pnpm v10.26+. These must be replaced with actual booleans based on whether those packages should be allowed to run postinstall scripts.

---

### 🟡 Medium

#### 2. Custom version comparison is naive and duplicates existing library

**File:** `src/core/react-query/plugin-package/helpers.ts`

The `compareVersionStrings` helper splits on `.` and parses segments with `Number.parseInt`, which silently ignores pre-release suffixes and non-numeric characters. Two versions like `1.0.0-alpha` and `1.0.0-beta` would both parse as `[1, 0, 0]` and compare as equal, potentially causing `IsUpdateAvailable` to be missed.

The project already depends on and uses `semver` in `src/pages/SentryErrorBoundaryWrapper.tsx`, `src/hooks/useIsFeatureSupported.ts`, and `src/components/Layout/TopNav.tsx`. You should use `semver.compare` or `semver.gt` instead of a hand-rolled parser.

#### 3. Broken plugin thumbnails show browser fallback instead of placeholder

**File:** `src/components/Settings/PluginManagement/InstalledPluginsPanel.tsx`

Unlike `CatalogItem.tsx`, which gracefully handles thumbnail load failures via `onError` + `imageFailed` state, this image has no error handler. If the thumbnail URL 404s, the user sees a broken image icon. Add `onError` handling to match the catalog behavior.

#### 4. `groupInstalledPlugins` uses quadratic-time object spreading

**File:** `src/core/react-query/plugin-package/helpers.ts`

Each iteration copies all existing keys into a new object and copies the existing array into a new array. This is O(n²). While plugin counts are likely small enough that it won't matter in practice, this pattern is unnecessarily inefficient and should be replaced with direct mutation of the accumulator (the reducer is local and safe to mutate).

---

### 🟢 Low / Style

#### 5. Inconsistent hook import style

**Files:** `src/components/Settings/PluginManagement/RepositoryPanel.tsx`, `src/components/Settings/PluginManagement/PluginUpdatesPanel.tsx`

`RepositoryPanel.tsx` imports `useState` destructured but then uses `React.useState` for one call and `useState` for another. `PluginUpdatesPanel.tsx` uses `React.useState` without a destructured import. The codebase overwhelmingly uses destructured imports (`import { useState } from 'react'`). Pick one style and use it consistently.

harshithmohan:
Not sure if the comments are valid, I just wanted to use my opencode go quota 😂

krbrs:
everything except the logo size is currently working well from the tests and reviews; i would re-model the interface eventually in another PR to fit the 16:9 logo if a smaller version is not feasible. Also settings for auto updates are currently not wired in because of testing

src/core/types/api/plugin-package.ts
```diff
@@ -0,0 +1,70 @@
...
+  IsCompatible: boolean;
+};
+
+export type PluginPackageArchiveType = {
+  RuntimeIdentifier: string;
+  AbstractionVersion: string;
+  ArchiveUrl: string;
+  ArchiveChecksum: string;
+  IsCompatible: boolean;
+};
```
harshithmohan: Looks like a duplicate of `PluginPackageArchiveType`?

src/core/types/api/plugin-package.ts
```diff
@@ -0,0 +1,70 @@
+import type { PluginInfoType } from '@/core/types/api/plugin';
+
+export type PackageRepositoryInfoType = {
+  ID: string;
+  Name: string;
+  Url: string;
+  LastFetchedAt?: string | null;
```
harshithmohan: If you already used `?` to make it optional, `| null` is not required. Same thing for all the types defined in this file

src/pages/settings/tabs/PluginsSettings.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { usePluginPackagesQuery } from '@/core/react-query/plugin-package/queries';
+
+const sections = [
+  { label: 'Installed', value: 'installed' },
+  { label: 'Browse', value: 'browse' },
+  { label: 'Updates', value: 'updates' },
+  { label: 'Repositories', value: 'repositories' },
+];
+
+const PluginsSettings = () => {
```
harshithmohan: Name it `PluginManagementSettings`

src/pages/settings/tabs/PluginsSettings.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const [section, setSection] = useState('installed');
```
harshithmohan: Try to put this in the url path instead of the state so it stays over refreshes

src/core/react-query/plugin-package/queries.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+export const usePluginPackagesQuery = (filters: PluginPackageListFilters) =>
+  useQuery<ListResultType<PackageInfoType>>({
+    queryKey: ['plugin-package', 'list', filters],
+    queryFn: () => axios.get('Plugin/Package', { params: filters }),
+    placeholderData: previousData => previousData,
+  });
+
+export const usePluginPackageVersionsQuery = (packageId: string | undefined) =>
```
harshithmohan:
```suggestion
export const usePluginPackageVersionsQuery = (packageId?: string) =>
```

src/core/react-query/plugin-package/queries.ts
```diff
@@ -0,0 +1,38 @@
...
+  useQuery<PackageInfoType[]>({
+    queryKey: ['plugin-package', 'versions', packageId],
+    queryFn: () => axios.get(`Plugin/Package/${packageId}`),
+    enabled: !!packageId,
+  });
+
+export const usePluginPackageManifestQuery = (packageId: string | undefined) =>
```
harshithmohan:
```suggestion
export const usePluginPackageManifestQuery = (packageId?: string) =>
```

src/core/react-query/plugin-package/queries.ts
```diff
@@ -0,0 +1,38 @@
...
+  useQuery<PackageManifestInfoType>({
+    queryKey: ['plugin-package', 'manifest', packageId],
```
harshithmohan: Swap packageId and 'manifest'

src/core/types/api/plugin-package.ts
```diff
@@ -0,0 +1,70 @@
...
+export type AddPackageRepositoryBodyType = {
+  name: string;
+  url: string;
+  staleTime?: string;
+};
+
+export type CheckForUpdatesBodyType = {
+  forceSync?: boolean;
+  performUpgrade?: boolean;
+};
```
harshithmohan: These two are request types and should be in `react-queryplugin-package/types.ts`. Also should be named `RequestType` and not `BodyType`

src/core/react-query/plugin-package/mutations.ts
```diff
@@ -0,0 +1,70 @@
...
+      axios.post(`Plugin/Package/Repository/${repositoryId}/Sync`, undefined, { params: { forceSync } }),
+    onSuccess: () => {
+      invalidateQueries(['plugin-package', 'repositories']);
+      invalidateQueries(['plugin-package', 'list']);
+    },
+  });
+
+export const useSyncAllPluginPackageRepositoriesMutation = () =>
+  useMutation({
+    mutationFn: ({ forceSync = true }: { forceSync?: boolean } = {}) =>
```
harshithmohan: `forceSync` is just a single value so you don't need to pass it as an object

src/core/react-query/plugin-package/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  Thumbnail?: PackageManifestInfoType['Thumbnail'];
+  LastFetchedAt: string;
+  Plugin?: PluginInfoType | null;
+  Releases: PluginPackageCatalogReleaseType[];
+  HasCompatibleInstallOption: boolean;
+  HasInstalledVersion: boolean;
+  HasUpdateAvailable: boolean;
+};
+
+export type PackageInstallMutationArgs = {
```
harshithmohan:
```suggestion
export type PackageInstallRequestType = {
```

src/core/react-query/plugin-package/queries.ts
```diff
@@ -0,0 +1,38 @@
...
+  useQuery<PackageRepositoryInfoType[]>({
+    queryKey: ['plugin-package', 'repositories'],
+    queryFn: () => axios.get('Plugin/Package/Repository'),
+  });
+
+export const usePluginPackagesQuery = (filters: PluginPackageListFilters) =>
+  useQuery<ListResultType<PackageInfoType>>({
+    queryKey: ['plugin-package', 'list', filters],
+    queryFn: () => axios.get('Plugin/Package', { params: filters }),
+    placeholderData: previousData => previousData,
```
harshithmohan: What is the placeholderData for?

src/core/react-query/plugin/queries.ts
```diff
@@ -0,0 +1,20 @@
...
+import { axios } from '@/core/axios';
+
+import type { PluginListFilters } from '@/core/react-query/plugin/types';
+import type { PluginInfoType } from '@/core/types/api/plugin';
+
+export const usePluginsQuery = (filters: PluginListFilters = {}) =>
+  useQuery<PluginInfoType[]>({
+    queryKey: ['plugin', 'list', filters],
+    queryFn: () => axios.get('Plugin', { params: filters }),
+    placeholderData: previousData => previousData,
```
harshithmohan: What is the placeholderData for?

src/core/react-query/plugin/queries.ts
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+export const usePluginVersionsQuery = (pluginId: string | undefined) =>
```
harshithmohan:
```suggestion
export const usePluginVersionsQuery = (pluginId?: string) =>
```

src/core/react-query/plugin/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  pluginId: string;
+  isEnabled: boolean;
+  pluginVersion?: string;
+};
+
+export type DeletePluginMutationArgs = {
+  pluginId: string;
+  purgeConfiguration?: boolean;
+  pluginVersion?: string;
+};
```
harshithmohan: `RequestType` not `MutationArgs`

AGENTS.md
```diff
@@ -87,3 +87,8 @@ This project uses the **React Compiler** (via `@rolldown/plugin-babel`). The com
 - Do not add explicit type annotations where TS inference is sufficient.
 - Treat changes to `src/core/axios.ts`, `src/core/store.ts`, and auth-related logic with extra scrutiny.
 - If you modify files, styles, structures, configurations, or workflows mentioned in this file, update the corresponding `AGENTS.md` sections to keep them accurate.
+
+## Specifications
+
+Implementation specs and planning artifacts may exist under `specs/`.
+Use them as historical/reference material when relevant, but prefer the current source code and API contracts as the source of truth.
```
harshithmohan, line 103: Do we need this?

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  query: string;
+};
+
+const RepositoryPanel = ({ query }: Props) => {
+  const repositoriesQuery = usePluginPackageRepositoriesQuery();
+  const { mutate: deleteRepository, status: deleteRepositoryStatus, variables: deletingId } =
+    useDeletePluginPackageRepositoryMutation();
+  const { mutate: syncRepository, status: syncRepositoryStatus, variables: syncingArgs } =
+    useSyncPluginPackageRepositoryMutation();
+  const { mutate: syncAll, status: syncAllStatus } = useSyncAllPluginPackageRepositoriesMutation();
```
harshithmohan: You can destructure `isPending` and use it instead of getting status and comparing it with `pending`

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [repositoryToDelete, setRepositoryToDelete] = React.useState<RepositoryDeleteStateType | null>(null);
```
harshithmohan: Import `useState` directly and use it, same for `useMemo`

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [isFormCollapsed, setIsFormCollapsed] = useState(true); // Start collapsed by default
+  const normalizedQuery = query.trim().toLocaleLowerCase();
```
harshithmohan: normalize it in the parent component when it's being set, not here

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+            buttonSize="small"
+            onClick={() => setIsFormCollapsed(!isFormCollapsed)}
+          >
+            Add Repository
+          </Button>
+        </div>
+      )}
+
+      {/* Show form only when appropriate */}
+      {shouldShowForm && <RepositoryForm />}
```
harshithmohan: The form should be a modal, not inline

harshithmohan: If you do that, you don't need to worry about hiding it when search query exists. The add repo button can just be always visible

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+    useDeletePluginPackageRepositoryMutation();
+  const { mutate: syncRepository, status: syncRepositoryStatus, variables: syncingArgs } =
+    useSyncPluginPackageRepositoryMutation();
+  const { mutate: syncAll, status: syncAllStatus } = useSyncAllPluginPackageRepositoriesMutation();
+  const [repositoryToDelete, setRepositoryToDelete] = React.useState<RepositoryDeleteStateType | null>(null);
+  const [isFormCollapsed, setIsFormCollapsed] = useState(true); // Start collapsed by default
+  const normalizedQuery = query.trim().toLocaleLowerCase();
+  const isSearching = normalizedQuery.length > 0;
+  const repositories = React.useMemo(() => {
+    if (!normalizedQuery) {
```
harshithmohan: You can just add this as the first line so you don't need to check with optional chaining every time

`if (!repositoryQuery.data) return [];`

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+            const isDeleting = deleteRepositoryStatus === 'pending' && deletingId === repository.ID;
+
+            return (
+              <div key={repository.ID} className="rounded-lg border border-panel-border bg-panel-background-alt p-4">
+                <div className="mb-2 flex flex-col gap-3 sm:flex-row sm:items-start sm:justify-between sm:gap-x-3">
+                  <div>
+                    <div className="font-semibold">{repository.Name}</div>
+                    <div className="text-sm opacity-80">{repository.Url}</div>
+                  </div>
+                  {isLocalRepository && (
```
harshithmohan: It's filtered out anyway....

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+                    <span className="rounded-lg border border-panel-border px-2 py-1 text-xs">Local</span>
+                  )}
+                </div>
+                <div className="mb-3 flex flex-wrap gap-x-1 text-sm opacity-65">
+                  <span>Last sync:</span>
+                  <span>
+                    {repository.LastFetchedAt ? new Date(repository.LastFetchedAt).toLocaleString() : 'Never'}
```
harshithmohan: use dayjs

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  </span>
+                </div>
+                {repository.StaleTime && (
+                  <div className="mb-3 flex flex-wrap gap-x-1 text-sm opacity-65">
+                    <span>Refresh interval:</span>
+                    <span>{repository.StaleTime}</span>
```
harshithmohan: What is this? Is it in hours?

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+      {/* Show form only when appropriate */}
+      {shouldShowForm && <RepositoryForm />}
+
+      <div className="rounded-lg border border-panel-border bg-panel-input p-4">
+        <div className="mb-4 flex flex-col gap-y-1">
+          <div className="text-lg font-semibold">Configured Repositories</div>
+          {isSearching && <div className="text-sm opacity-70">Clear the search field to add a new repository.</div>}
+        </div>
+        {repositoriesQuery.data?.length === 0 && (
```
harshithmohan: length is 1 when no repos are configured because of the local repo

src/components/Settings/PluginManagement/PluginUpdatesPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+  entries: PluginPackageCatalogEntryType[];
+  isLoading: boolean;
+  updates: PluginUpdateSummaryType[];
+};
+
+const PluginUpdatesPanel = ({ entries, isLoading, updates }: Props) => {
+  const { mutate: checkUpdates, status } = useCheckPluginPackageUpdatesMutation();
+  const [selectedPackageId, setSelectedPackageId] = React.useState<string>();
+  const selectedEntry = entries.find(entry => entry.PackageID === selectedPackageId);
+  const updatesContent = (() => {
```
harshithmohan: This can be a separate component if you want but shouldn't be an IIFE

src/components/Settings/PluginManagement/InstallPluginDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+            disabled={!canInstall}
+          >
+            Install
+          </Button>
+        </div>
+      }
+    >
+      {!entry
+        ? null
+        : (
```
harshithmohan:
```suggestion
      {entry && (
```

src/components/Settings/PluginManagement/InstalledPluginsPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+};
+
+const Badge = ({ children, className }: BadgeProps) => (
+  <span className={`rounded-lg px-2.5 py-1 text-xs font-medium ${className ?? ''}`.trim()}>
+    {children}
+  </span>
+);
+
+const InstalledPluginsPanel = ({ groupedPackages, groupedPlugins }: Props) => {
+  const [expandedIds, setExpandedIds] = useState<Record<string, boolean>>({});
```
harshithmohan: Wouldn't it be ok if we just allowed one to expand at a time?

src/components/Settings/PluginManagement/InstalledPluginsPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+              >
+                <div className="w-full shrink-0 lg:w-44 lg:flex-none">
+                  {thumbnailUrl
+                    ? (
+                      <img
+                        src={thumbnailUrl}
+                        alt={plugin.Name}
+                        className="block aspect-video w-full rounded-lg border border-panel-border bg-panel-background-alt object-contain"
+                        onError={() => {
+                          if (!thumbnailUrl) return;
```
harshithmohan: This line does nothing, it won't be rendered is `thumbnailUrl` doesn't exist

src/components/Settings/PluginManagement/InstalledPluginsPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+        ].filter(candidate => !!candidate && !failedThumbnailUrls[candidate]);
+        const thumbnailUrl = thumbnailCandidates[0];
+        const expanded = !!expandedIds[pluginId];
+        const updateEntry = groupedPackages.find(entry => entry.Plugin?.ID === pluginId && entry.HasUpdateAvailable);
+        const hasReadOnlyVersions = plugins.every(
+          installedPlugin => installedPlugin.RestartPending || !installedPlugin.CanUninstall,
+        );
+
+        return (
+          <article
```
harshithmohan: article? 🤔

src/components/Settings/PluginManagement/InstalledPluginVersions.tsx
```diff
@@ -0,0 +1,223 @@
...
+                {plugin.CanLoad && (
+                  <Badge className="border border-panel-border bg-panel-input text-inherit">
+                    Compatible
+                  </Badge>
+                )}
+              </div>
+
+              <div className="flex flex-wrap items-center gap-x-2 gap-y-1 text-sm">
+                <span>Installed:</span>
+                <span>{new Date(plugin.InstalledAt).toLocaleDateString()}</span>
```
harshithmohan: Use dayjs

harshithmohan (changes requested):

AGENTS.md
krbrs: This is automatically added by Spec-Kit; I already made it more general. If I remove it and run an AI over it, it will not look into the Spec-Kit specs folder to work with. That's why it's currently added. Normaly it will include some block around it and only point it to the current project.

src/components/Settings/PluginManagement/RepositoryPanel.tsx
krbrs: yeah, it's in format 12:00:00 - that's directly taken from the swagger openapi specs

src/components/Settings/PluginManagement/PluginManagement.utils.ts
```diff
@@ -0,0 +1,3 @@
+import { dayjs } from '@/core/util';
+
+export const formatPluginDate = (date: string) => dayjs(date).format('D MMMM YYYY');
```
harshithmohan: I was wrong, we apparently use either `MMMM Do, YYYY` or `MMM DD, YYYY`.
So use any combination of month and date out of those

MMMM is full month name, MMM is three letter name
`Do` is date with the suffix eg., st, th
`DD` is just the date

src/pages/settings/SettingsPage.tsx
```diff
@@ -64,6 +65,8 @@ const SettingsPage = () => {
   const [debouncedUnsavedChanges] = useDebounceValue(unsavedChanges, 100);
 
   const isSpecialPage = useMemo(() => {
+    if (pathname.includes('/settings/plugins')) return true;
```
harshithmohan: Combine this with line 72, also maybe change the plugin config page to be `plugins` instead of `plugin`

harshithmohan: Wait that might break things.. just change this one to `plugin-management`

src/pages/settings/tabs/PluginManagementSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const isValidSection = (section?: string): section is typeof sections[number]['value'] =>
+  !!section && sections.some(item => item.value === section);
+
+const PluginManagementSettings = () => {
+  const navigate = useNavigateVoid();
+  const { section } = useParams();
+  const [query, setQuery] = useState('');
+  const [debouncedQuery] = useDebounceValue(query, 300);
+  const trimmedQuery = debouncedQuery.trim();
```
harshithmohan: Just trim on the first line... `useDebounceValue(query.trim(), 300)`

src/pages/settings/tabs/PluginManagementSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const selectedSection = isValidSection(section) ? section : defaultSection;
```
harshithmohan: If it's not invalid, we should redirect them to the default section, not show the default section for any string

harshithmohan: Which is already being done below.. so this line itself doesn't matter

src/pages/settings/tabs/PluginManagementSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {selectedSection === 'repositories' && <RepositoryPanel query={trimmedQuery} />}
+      {selectedSection === 'browse' && <CatalogPanel query={trimmedQuery} />}
+      {selectedSection === 'installed' && <InstalledPluginsPanel query={trimmedQuery} />}
+      {selectedSection === 'updates' && <PluginUpdatesPanel query={trimmedQuery} />}
+    </>
+  );
+
+  if (!isValidSection(section)) {
+    content = <Navigate replace to={`/webui/settings/plugins/${defaultSection}`} />;
+  }
```
harshithmohan: Why is being stored in content?

Put the `if` above and inside just return the Navigate component

harshithmohan: Also the `defaultSection` var won't be needed anymore, it will only be used at once place

src/pages/settings/tabs/PluginManagementSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div className="max-w-3xl">
+          Manage plugin repositories, browse plugin packages, review installed plugins, and manually apply updates.
+        </div>
+      </div>
+
+      <div className="border-b border-panel-border" />
+
+      <MultiStateButton
+        activeState={selectedSection}
+        onStateChange={state => navigate(`/webui/settings/plugins/${state}`)}
```
harshithmohan: Can just write `../${state}`
Same for Navigate above

src/components/Settings/PluginManagement/RepositoryForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+  useEffect(() => {
+    if (!show) return;
+    reset();
+  }, [show]);
+
+  const handleSubmit = () => {
+    if (!name.trim() || !url.trim()) {
+      toast.warning('Repository details required', 'Provide both a repository name and manifest URL.');
+      return;
+    }
```
harshithmohan: We can instead disable the save button when either of those is not filled and show a tooltip

src/components/Settings/PluginManagement/RepositoryForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    addRepository({ name: name.trim(), url: url.trim() }, {
+      onSuccess: () => {
+        toast.success('Repository added', `${name.trim()} has been added.`);
+        reset();
+        onClose();
+      },
+    });
```
harshithmohan, line 51: No error toast? 🤔

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+                    buttonType="danger"
+                    buttonSize="small"
+                    onClick={() => setRepositoryToDelete({ ID: repository.ID, Name: repository.Name })}
+                    loading={isDeleting}
+                  >
+                    Remove
+                  </Button>
+                </div>
+              </div>
+            );
```
harshithmohan: Move this to a separate component in the same file. That would also move both the mutation hooks into that component and you wouldn't need to check isSyncing or isDeleting using the variables anymore

harshithmohan: Also write a handler function for sync

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+            })}
+          loading={isSyncAllPending}
+        >
+          Sync All
+        </Button>
+      </div>
+
+      <div className="rounded-lg border border-panel-border bg-panel-input p-4">
+        <div className="mb-4 text-lg font-semibold">Configured Repositories</div>
+        {repositoriesQuery.data && repositories.length === 0 && !query && (
```
harshithmohan: ~~Put `query`/`!query` at the start for easier distinction.~~
Otherwise both can be combined. Main condition as `repositoriesQuery.data && repositories.length === 0` and only change the text based on `query` with ternary

harshithmohan: Is `repositoriesQuery.data` even needed in the condition?
The code won't reach here if it's pending or returns an error

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (!repositoriesQuery.data) return [];
+    if (!query) return repositoriesQuery.data;
+
+    const matchesQuery = (value: string) => value.toLowerCase().includes(query.toLowerCase());
+
+    return repositoriesQuery.data.filter(repository => [repository.Name, repository.Url].some(matchesQuery));
+  }, [query, repositoriesQuery.data]);
+  const retryRepositories = () => {
+    repositoriesQuery.refetch().catch(console.error);
+  };
```
harshithmohan: Move this directly to the onClick handler

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <Button
+          buttonType="secondary"
+          buttonSize="normal"
+          onClick={() =>
+            syncAll(true, {
+              onSuccess: () => toast.success('Repositories synchronized'),
+              onError: () => toast.error('Failed to synchronize repositories'),
+            })}
+          loading={isSyncAllPending}
+        >
```
harshithmohan: onClick is too big, move to handler

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <Icon path={mdiLoading} spin size={4} />
+      </div>
+    );
+  }
+
+  if (repositoriesQuery.isError) {
+    return (
+      <div className="rounded-lg border border-panel-border bg-panel-input p-6">
+        <div className="text-lg font-semibold">Repositories unavailable</div>
+        <div className="mt-2 opacity-80">
```
harshithmohan: I think we use opacity-65 at other places in the existing codebase

src/components/Settings/PluginManagement/RepositoryPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+const RepositoryPanel = ({ query }: Props) => {
+  const repositoriesQuery = usePluginPackageRepositoriesQuery();
+  const { isPending: isDeletePending, mutate: deleteRepository, variables: deletingId } =
+    useDeletePluginPackageRepositoryMutation();
+  const { isPending: isSyncPending, mutate: syncRepository, variables: syncingRepositoryId } =
+    useSyncPluginPackageRepositoryMutation();
+  const { isPending: isSyncAllPending, mutate: syncAll } = useSyncAllPluginPackageRepositoriesMutation();
+  const [isAddRepositoryOpen, setIsAddRepositoryOpen] = useState(false);
+  const [repositoryToDelete, setRepositoryToDelete] = useState<RepositoryDeleteStateType | null>(null);
```
harshithmohan: Why `| null`? If you remove and that the default value, it will be undefined which would work the same way

src/core/react-query/plugin-package/mutations.ts
```diff
@@ -0,0 +1,73 @@
...
+  CheckForUpdatesRequestType,
+  PackageInstallRequestType,
+} from '@/core/react-query/plugin-package/types';
+
+export const useAddPluginPackageRepositoryMutation = () =>
+  useMutation({
+    mutationFn: (body: AddPackageRepositoryRequestType) => axios.post('Plugin/Package/Repository', body),
+    onSuccess: () => {
+      invalidateQueries(['plugin-package', 'repositories']);
+      invalidateQueries(['plugin-package', 'list']);
```
harshithmohan: I remember that these two lines were part of a function which could be re-used by the 4 mutations, where did that go? 😔 
Add that in helpers and just use that for all 4

src/core/react-query/plugin-package/mutations.ts
```diff
@@ -0,0 +1,73 @@
...
+    },
+  });
+
+export const useSyncAllPluginPackageRepositoriesMutation = () =>
+  useMutation({
+    mutationFn: (forceSync?: boolean) =>
+      axios.post('Plugin/Package/Repository/Sync', undefined, { params: { forceSync: forceSync ?? true } }),
```
harshithmohan:
```suggestion
    mutationFn: (forceSync = true) =>
      axios.post('Plugin/Package/Repository/Sync', undefined, { params: { forceSync } }),
```

src/core/react-query/plugin-package/mutations.ts
```diff
@@ -0,0 +1,73 @@
...
+export const useInstallPluginPackageMutation = () =>
+  useMutation({
+    mutationFn: ({ abstractionVersion, packageId, releaseVersion, runtimeIdentifier }: PackageInstallRequestType) =>
+      axios.post(`Plugin/Package/${packageId}/Install`, undefined, {
+        params: {
+          abstractionVersion,
+          releaseVersion,
+          runtimeIdentifier,
+        },
+      }),
```
harshithmohan:
```suggestion
    mutationFn: ({ packageId, ...params }: PackageInstallRequestType) =>
      axios.post(`Plugin/Package/${packageId}/Install`, undefined, { params }),
```

src/core/react-query/plugin-package/queries.ts
```diff
@@ -0,0 +1,39 @@
...
+    queryFn: () => axios.get(`Plugin/Package/${packageId}`),
+    enabled: !!packageId,
+  });
+
+export const usePluginPackageManifestQuery = (packageId?: string) =>
+  useQuery<PackageManifestInfoType>({
+    queryKey: ['plugin-package', packageId, 'manifest'],
+    queryFn: () => axios.get(`Plugin/Package/${packageId}/Manifest`),
+    enabled: !!packageId,
+  });
```
harshithmohan: If these are not required in the future, remove them, they are unused.
If they are, `packageId` shouldn't be optional and there should an enabled flag

src/core/react-query/plugin-package/types.ts
```diff
@@ -0,0 +1,74 @@
...
+  allowSync?: boolean;
+  forceSyncNow?: boolean;
+};
+
+export type PluginPackageCatalogArchiveType = PackageArchiveInfoType & {
+  IsCompatible: boolean;
+  IsInstalled: boolean;
+};
+
+export type PluginPackageCatalogReleaseType = Omit<PackageReleaseInfoType, 'Archives'> & {
```
harshithmohan, line 17: You don't need the `Omit`, the below `Archives` will override it anyway

src/core/react-query/plugin-package/types.ts
```diff
@@ -0,0 +1,74 @@
...
+  IsUpdateAvailable: boolean;
+};
+
+export type PluginPackageCatalogEntryType = {
+  PackageID: string;
+  Name: string;
+  Overview: string;
+  Authors: string;
+  Tags: string[];
+  Thumbnail?: PackageManifestInfoType['Thumbnail'];
```
harshithmohan: Separate the Thumbnail type and use it

src/core/react-query/plugin-package/types.ts
```diff
@@ -0,0 +1,74 @@
...
+  url: string;
+  staleTime?: string;
+};
+
+export type CheckForUpdatesRequestType = {
+  forceSync?: boolean;
+  performUpgrade?: boolean;
+};
+
+export type PackageManifestLookup = Record<string, PluginPackageCatalogEntryType>;
```
harshithmohan: Looks unused

src/core/react-query/plugin-package/types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type PluginUpdateSummaryType = {
+  PackageID: string;
+  Name: string;
+  CurrentVersion: string;
+  LatestVersion: string;
+  Release: PluginPackageCatalogReleaseType;
+};
+
+export type PackageVersionsLookupType = Record<string, PackageInfoType[]>;
```
harshithmohan: Looks unused

src/core/react-query/plugin/queries.ts
```diff
@@ -3,7 +3,22 @@ import { useQuery } from '@tanstack/react-query';
 import { axios } from '@/core/axios';
 import { STALE_TIME } from '@/core/util';
 
-import type { PluginPageType, SharedPluginPageType } from '@/core/types/api/plugin';
+import type { PluginListFilters } from '@/core/react-query/plugin/types';
+import type { PluginInfoType, PluginPageType, SharedPluginPageType } from '@/core/types/api/plugin';
+
+export const usePluginsQuery = (filters: PluginListFilters = {}, enabled = true) =>
```
harshithmohan: `filters` doesn't need a default value

src/core/react-query/plugin/queries.ts
```diff
@@ -3,7 +3,22 @@ import { useQuery } from '@tanstack/react-query';
...
+    queryFn: () => axios.get('Plugin', { params: filters }),
+    enabled,
+  });
+
+export const usePluginVersionsQuery = (pluginId?: string) =>
+  useQuery<PluginInfoType[]>({
+    queryKey: ['plugin', 'versions', pluginId],
+    queryFn: () => axios.get(`Plugin/${pluginId}/All`),
+    enabled: !!pluginId,
+  });
```
harshithmohan: Remove if not using or add separate enabled flag

src/core/react-query/plugin/types.ts
```diff
@@ -0,0 +1,29 @@
+import type { PluginInfoType } from '@/core/types/api/plugin';
+
+// Mirrors `Shoko.Server.API.v3.Models.Common.IncludeOnlyFilter`. The server
+// defaults each filter to `True`, which means "include everything"; `Only`
+// keeps just matching entries and `False` excludes them.
+export type PluginIncludeOnlyFilter = 'True' | 'False' | 'Only';
```
harshithmohan: Wouldn't `IncludeOnlyFilterEnum` from `react-query/series/types.ts` work for this?
If it does, move that enum to `react-query/types.ts` and use it

src/core/react-query/plugin/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  pluginVersion?: string;
+};
+
+export type DeletePluginRequestType = {
+  pluginId: string;
+  purgeConfiguration?: boolean;
+  pluginVersion?: string;
+};
+
+export type PluginGroupsType = Record<string, PluginInfoType[]>;
```
harshithmohan: `Record` can be directly used, you don't need to declare a type for just using it at one place

src/core/react-query/plugin/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  active?: PluginIncludeOnlyFilter;
+  installed?: PluginIncludeOnlyFilter;
+  enabled?: PluginIncludeOnlyFilter;
+  restartPending?: PluginIncludeOnlyFilter;
+  allVersions?: boolean;
+};
+
+export type UpdatePluginRequestType = {
+  pluginId: string;
+  IsEnabled?: boolean;
```
harshithmohan: Should be `isEnabled`

src/core/react-query/plugin/mutations.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { axios } from '@/core/axios';
+import { invalidateQueries } from '@/core/react-query/queryClient';
+
+import type { DeletePluginRequestType, UpdatePluginRequestType } from '@/core/react-query/plugin/types';
+
+const invalidatePluginQueries = () => {
+  invalidateQueries(['plugin']);
+  invalidateQueries(['plugin-package', 'list']);
+};
```
harshithmohan: Oh, it was here the function existed. Move this to helpers

src/components/Settings/PluginManagement/PluginUpdatesPanel.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const packagesQuery = usePluginPackagesQuery({
+    allowSync: false,
+    onlyCompatible: false,
+    onlyLatest: false,
+    pageSize: 0,
+  });
+  const entries = useMemo(
+    () => groupPluginPackages(packagesQuery.data?.List ?? []),
+    [packagesQuery.data?.List],
+  );
```
harshithmohan: `packagesQuery.data` is only like this, So instead of calling `groupPluginPackages` here, it can be used with the `select` parameter in the query to transform what is stored in the cache.
Then you won't need to call the group function at 3 places like it is called now

harshithmohan (changes requested):
There are a lot of places where tailwind inbuilt color classes are used such as `bg-orange` or `border-orange`. Those should be replaced by our theme colors

src/core/react-query/plugin/types.ts
krbrs: I checked this, but the existing enum is not reusable safely.

`IncludeOnlyFilterEnum` in `series/types.ts` sends lowercase values:

`'true' | 'false' | 'only'`

The plugin endpoint expects title-case values:

`'True' | 'False' | 'Only'`

So while they represent the same server-side concept, the wire values differ between controllers. Reusing one shared enum would send the wrong casing to one of the APIs unless we also add a conversion layer. I kept the plugin-specific type for now.

src/core/react-query/plugin-package/types.ts
krbrs: I tried removing the `Omit`, but TypeScript does not narrow the overridden `Archives` field cleanly in the intersection type. Keeping `Omit<PackageReleaseInfoType, 'Archives'>` makes the replacement explicit and preserves the expected `PluginPackageCatalogArchiveType[]` typing.

src/components/Settings/PluginManagement/RepositoryForm.tsx
krbrs: This has been addressed — `addRepository` now has an `onError` handler that shows `toast.error('Failed to add repository', ...)`.

src/core/react-query/plugin/types.ts
harshithmohan: Sounds like something we should complain about

krbrs: Yeah, agreed. I kept the frontend type separate for now so we don't paper over the mismatch, but the API casing should probably be normalized server-side. Once both endpoints accept/emit the same wire values, this can become a shared react-query type.

harshithmohan: Same enum is used server-side for all endpoints. It is not case sensitive, the enum converter handles it fine.

src/core/react-query/plugin-package/types.ts
harshithmohan: Ok then 😔

PR #1340, opened 2026-05-31
feat(settings): add plugin pages to settings sidebar with iframe embeds

revam:
Plugin pages are listed in the settings sidebar, grouped by plugin, with embeddable pages opening in an `<iframe/>` via the new new `/settings/plugin/:id/:index` route, and external pages opening in a new tab. External pages have an external link icon attched to them. Everything tested locally.

Example plugin page;
image

### DIFF
--- a/src/core/react-query/plugin/queries.ts
+++ b/src/core/react-query/plugin/queries.ts
@@ -1,6 +1,7 @@
 import { useQuery } from '@tanstack/react-query';
 
 import { axios } from '@/core/axios';
+import { STALE_TIME } from '@/core/util';
 
 import type { PluginPageType, SharedPluginPageType } from '@/core/types/api/plugin';
 
@@ -8,13 +9,13 @@ export const usePluginPagesQuery = () =>
   useQuery<PluginPageType[]>({
     queryKey: ['plugin', 'pages'],
     queryFn: () => axios.get('Plugin/Pages'),
-    staleTime: 86400 * 100000,
+    staleTime: STALE_TIME,
   });
 
 export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
   useQuery<SharedPluginPageType[]>({
     queryKey: ['plugin', 'pages', pluginId],
     queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
-    staleTime: 86400 * 100000,
+    staleTime: STALE_TIME,
     enabled,
   });
--- a/src/core/react-query/plugin/queries.ts
+++ b/src/core/react-query/plugin/queries.ts
@@ -2,7 +2,7 @@ import { useQuery } from '@tanstack/react-query';
 
 import { axios } from '@/core/axios';
 
-import type { PluginPageType } from '@/core/types/api/plugin';
+import type { PluginPageType, SharedPluginPageType } from '@/core/types/api/plugin';
 
 export const usePluginPagesQuery = () =>
   useQuery<PluginPageType[]>({
@@ -12,7 +12,7 @@ export const usePluginPagesQuery = () =>
   });
 
 export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
-  useQuery<PluginPageType[]>({
+  useQuery<SharedPluginPageType[]>({
     queryKey: ['plugin', 'pages', pluginId],
     queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
     staleTime: 86400 * 100000,
--- a/src/core/react-query/plugin/queries.ts
+++ b/src/core/react-query/plugin/queries.ts
@@ -1,0 +1,18 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { axios } from '@/core/axios';
+
+import type { PluginPageType } from '@/core/types/api/plugin';
+
+export const usePluginPagesQuery = () =>
+  useQuery<PluginPageType[]>({
+    queryKey: ['plugin', 'pages'],
+    queryFn: () => axios.get('Plugin/Pages'),
+  });
+
+export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
+  useQuery<PluginPageType[]>({
+    queryKey: ['plugin', 'pages', pluginId],
+    queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
+    enabled,
+  });
--- a/src/core/react-query/plugin/queries.ts
+++ b/src/core/react-query/plugin/queries.ts
@@ -8,11 +8,13 @@ export const usePluginPagesQuery = () =>
   useQuery<PluginPageType[]>({
     queryKey: ['plugin', 'pages'],
     queryFn: () => axios.get('Plugin/Pages'),
+    staleTime: 86400 * 100000,
   });
 
 export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
   useQuery<PluginPageType[]>({
     queryKey: ['plugin', 'pages', pluginId],
     queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
+    staleTime: 86400 * 100000,
     enabled,
   });
--- a/src/core/react-query/plugin/queries.ts
+++ b/src/core/react-query/plugin/queries.ts
@@ -6,14 +6,14 @@ import { STALE_TIME } from '@/core/util';
 import type { PluginPageType, SharedPluginPageType } from '@/core/types/api/plugin';
 
 export const usePluginPagesQuery = () =>
-  useQuery<PluginPageType[]>({
+  useQuery<SharedPluginPageType[]>({
     queryKey: ['plugin', 'pages'],
     queryFn: () => axios.get('Plugin/Pages'),
     staleTime: STALE_TIME,
   });
 
 export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
-  useQuery<SharedPluginPageType[]>({
+  useQuery<PluginPageType[]>({
     queryKey: ['plugin', 'pages', pluginId],
     queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
     staleTime: STALE_TIME,
--- a/src/core/types/api/plugin.ts
+++ b/src/core/types/api/plugin.ts
@@ -36,6 +36,7 @@ export type PluginInfoType = {
 };
 
 export type PluginPageType = {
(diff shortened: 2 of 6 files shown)


harshithmohan:
@hidden4003 What are we doing this?

hidden4003:
@harshithmohan I ran it through Claude
# Code Review: PR #1340 — Plugin Pages in Settings Sidebar

## Overview

Adds a plugin page section to the settings sidebar. Embeddable pages render in an iframe via `/settings/plugin/:pluginId/:pageIndex`; non-embeddable pages open in a new tab. Five files changed: two new query hooks, a new type, a new `PluginPageEmbed` component, and sidebar wiring in `SettingsPage`.

---

## 🔴 Critical: Page Index Mismatch

This is the most important issue.

**`SettingsPage.tsx`** generates the link index from `Plugin/Pages` (the global query):

```ts
group.pages.map((page, pageIdx) => (
  <NavLink to={`plugin/${group.id}/${pageIdx}`} ...>
```

**`PluginPageEmbed.tsx`** resolves that index from `Plugin/{pluginId}/Pages` (a per-plugin query):

```ts
const { data: pages } = usePluginPagesForPluginQuery(pluginId);
const page = pages?.[pageIndex];
```

If the two endpoints return pages in a different order — entirely plausible since they are separate API endpoints — the wrong page will be loaded silently. There is no correctness guarantee that index `0` in the global response equals index `0` in the per-plugin response.

**Fix:** Navigate by page name or a stable page ID rather than a positional index, e.g. `plugin/${group.id}/${encodeURIComponent(page.Name)}` and look up `pages?.find(p => p.Name === pageName)`. Or make `PluginPageEmbed` use the same global query (`usePluginPagesQuery`) and filter by `PluginInfo?.ID`, keeping the indexing consistent.

---

## 🔴 Security: `postMessage` Handler Missing Origin/Source Check

```ts
const handler = (event: MessageEvent) => {
  if (... data.type === 'shoko-plugin-resize' ... typeof data.height === 'number') {
    setIframeHeight(data.height);
  }
};
globalThis.addEventListener('message', handler);
```

Two problems:

1. **No `event.source` check.** Any page open in the browser that has a reference to this window can send `{ type: 'shoko-plugin-resize', height: 0 }` and collapse the iframe to invisible, or send `height: 999999` to break the layout. Minimum fix: validate `event.source === iframeRef.current?.contentWindow`.

2. **No height clamping.** Even with source validation a rogue plugin could send an absurd value. Add a clamp:
   ```ts
   const safeHeight = Math.max(50, Math.min(data.height, 8000));
   setIframeHeight(safeHeight);
   ```

---

## 🔴 Security: `<iframe>` Has No `sandbox` Attribute

```tsx
<iframe ref={iframeRef} src={page.Url} title={page.Name} ... />
```

Without sandboxing, a same-origin plugin page can access and mutate the parent window's DOM, its React state, and the Shoko auth token. Even cross-origin pages can attempt top-frame navigation. If plugins can originate from third-party developers, this is a real attack surface.

**Recommendation:** Add `sandbox="allow-scripts allow-forms allow-same-origin"` as a baseline and document the trust model. Note that `allow-same-origin` + `allow-scripts` together on a same-origin iframe still allows escaping the sandbox — if same-origin plugin pages need to be supported, make this explicit in a comment so future maintainers understand the trade-off.

---

## 🟡 Bug: `pluginId` Defaults to Empty String

```ts
const { pageIndex: pageIndexStr = '0', pluginId = '' } = useParams();
// ...
const { data: pages, isPending } = usePluginPagesForPluginQuery(pluginId);
```

The default `pluginId = ''` means the hook fires a request to `Plugin//Pages` if params are ever absent.

**Fix:** Pass `pluginId !== ''` as the `enabled` argument:

```ts
const { data: pages, isPending } = usePluginPagesForPluginQuery(pluginId, pluginId !== '');
```

---

## 🟡 Bug: React Key Collision on `page.Name`

```tsx
group.pages.map((page, pageIdx) => (
  page.CanEmbed ? <NavLink key={page.Name} ...> : <a key={page.Name} ...>
))
```

If a plugin has two pages with the same name, React will silently reuse DOM nodes causing rendering bugs.

**Fix:** Use a composite key:

```tsx
key={`${group.id}-${pageIdx}`}
```

---

## 🟢 Minor Issues

### Unnecessary `pages` Dependency in ResizeObserver Effect

```ts
useEffect(() => {
  // setupObserver doesn't use `pages` at all
}, [pluginId, pageIndex, pages]); // `pages` is not used inside
```

`pages` is listed as a dependency but never read inside the effect body. This causes the ResizeObserver to disconnect and re-register on every background refetch. Remove it — `[pluginId, pageIndex]` is sufficient.

---

### `isSpecialPage` — Inconsistent Path Matching

```ts
if (pathname.includes('settings/dynamic/') || pathname.includes('settings/plugin')) return true;
```

`settings/dynamic/` requires a trailing slash but `settings/plugin` does not. A hypothetical future route `settings/plugin-foo` would incorrectly match.

**Fix:** Use `settings/plugin/` for consistency.

---

### `_unknown` Group Visible to Users

```ts
const pluginId = page.PluginInfo?.ID ?? '_unknown';
```

Pages where `PluginInfo` is absent all land in a single `_unknown` group that is rendered in the sidebar. Either filter these out entirely or use a user-readable fallback label like `"Unknown Plugin"`.

---

### Magic `+ 48` Height Offset

```ts
const height = doc.body.scrollHeight + 48; // We have a 3rem margin we need to negate.
```

`48px = 3rem` holds only for `font-size: 16px`. If the root font size changes the iframe will be sized incorrectly. Consider computing the offset from the wrapper element's actual computed style rather than a hardcoded pixel value.

---

## What's Done Well

- `rel="noopener noreferrer"` on external links is correct.
- The `try/catch` for cross-origin `contentDocument` access is the right approach and fails gracefully.
- Splitting the two query hooks (`usePluginPagesQuery` / `usePluginPagesForPluginQuery`) with distinct query keys is clean.
- The `pluginGroups` memo correctly avoids re-computing on unrelated renders.
- The height-reset effect on `[pluginId, pageIndex]` is correct and prevents stale dimensions when navigating between pages.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 Critical | Page index mismatch between global and per-plugin queries |
| 2 | 🔴 High | `postMessage` handler has no `event.source` / origin check |
| 3 | 🔴 High | No height clamping on `postMessage` value |
| 4 | 🔴 High | `<iframe>` has no `sandbox` attribute |
| 5 | 🟡 Medium | `pluginId = ''` fires a bad API request |
| 6 | 🟡 Medium | `page.Name` as React key can collide |
| 7 | 🟢 Low | `pages` in ResizeObserver dep array is unnecessary |
| 8 | 🟢 Low | `settings/plugin` missing trailing slash in `isSpecialPage` |
| 9 | 🟢 Low | `_unknown` group label visible to users |
| 10 | 🟢 Low | Hardcoded `+ 48` offset is fragile under non-default font sizes |

**Do not merge as-is.** Items 1–4 are blockers: the index mismatch is a correctness bug that will silently show the wrong page; the `postMessage` and sandbox issues are security concerns that should be resolved before the plugin API is available to third-party developers.

src/core/react-query/plugin/queries.ts
```diff
@@ -0,0 +1,20 @@
...
+    queryKey: ['plugin', 'pages'],
+    queryFn: () => axios.get('Plugin/Pages'),
+    staleTime: 86400 * 100000,
+  });
+
+export const usePluginPagesForPluginQuery = (pluginId: string, enabled = true) =>
+  useQuery<PluginPageType[]>({
+    queryKey: ['plugin', 'pages', pluginId],
+    queryFn: () => axios.get(`Plugin/${pluginId}/Pages`),
+    staleTime: 86400 * 100000,
```
harshithmohan: Can you just make this a constant in Util.ts so we can re-use it?

src/pages/settings/SettingsPage.tsx
```diff
@@ -41,6 +42,25 @@ const SettingsPage = () => {
   const settings = settingsQuery.data;
   const { isPending: settingsPatchPending, mutate: patchSettings } = usePatchSettingsMutation();
 
+  const { data: pluginPages } = usePluginPagesQuery();
+
+  const pluginGroups = useMemo(() => {
+    if (!pluginPages) return [];
+    const groups: { id: string, name: string, pages: typeof pluginPages }[] = [];
+    const seen = new Map<string, typeof pluginPages>();
```
harshithmohan: We know the type is `PluginPageType[]` so use that instead of `typeof pluginPages`

src/pages/settings/SettingsPage.tsx
```diff
@@ -41,6 +42,25 @@ const SettingsPage = () => {
   const settings = settingsQuery.data;
   const { isPending: settingsPatchPending, mutate: patchSettings } = usePatchSettingsMutation();
 
+  const { data: pluginPages } = usePluginPagesQuery();
```
harshithmohan:
```suggestion
  const pluginPages = usePluginPagesQuery().data;
```

For consistency

src/pages/settings/SettingsPage.tsx
harshithmohan: Brownie points if you can figure out why :pensive:

src/core/types/api/plugin.ts
```diff
@@ -5,3 +5,11 @@ export type PluginInfoType = {
   Description: string | null;
   CanUninstall: boolean;
 };
+
+export type PluginPageType = {
+  ID: string;
+  Name: string;
+  Url: string;
+  CanEmbed: boolean;
+  PluginInfo?: PluginInfoType;
```
hidden4003: ## 🟡 `PluginInfo` Typed as Optional but Treated as Required

```ts
// Type definition — still optional:
export type PluginPageType = {
  ID: string;
  Name: string;
  Url: string;
  CanEmbed: boolean;
  PluginInfo?: PluginInfoType;   // <-- optional
};

// Usage — non-null asserted:
const pluginId = page.PluginInfo!.ID;
const name     = pages[0].PluginInfo!.Name;
```

The commit message confirms the shared endpoint always returns `PluginInfo`. The type should reflect that: remove the `?` from `PluginInfo` in `PluginPageType`. Keeping it optional while asserting non-null everywhere gives a false sense of type safety and will produce a cryptic runtime crash rather than a compile error if the API ever violates the contract.

hidden4003 (changes requested):

hidden4003 (approved):

PR #1342, opened 2026-06-01
Migrate renamers to new system, add dynamic UI

harshithmohan:


src/core/types/api/configuration.ts
```diff
@@ -0,0 +1,115 @@
...
+  selectMultipleItems?: boolean;
+  selectType?: string;
+  showSaveAction?: boolean;
+  structure?: Record<string, string>;
+  visibility?: UiVisibilityType;
+};
+
+export type PropertySchemaType = {
+  $ref?: string;
+  additionalProperties?: PropertySchemaType;
```
hidden4003: this one is called `additionalProperties` which suggests it is a `Record<>` but it is typed as a single element, although you can keep nesting `additionalProperties` indefinitely I guess, but does it reflect what server sends?

src/components/Dynamic/DynamicField.tsx
```diff
@@ -0,0 +1,131 @@
...
+  propertySchema: PropertySchemaType;
+  value: unknown;
+};
+
+const DynamicField = ({ onChange, propertyName, propertySchema, value }: Props) => {
+  const { 'x-uiDefinition': uiDef } = propertySchema;
+  const { elementType } = uiDef;
+  const label = propertySchema.title ?? propertyName;
+
+  if (elementType === 'auto') {
```
hidden4003, line 28: This pattern will lead to a huge file if we implement the rest of the elements.

hidden4003 (approved):

hidden4003 (approved):
I hope for simpler schema in the future for the ui similar to this one we discussed, but for now it should be fine. Simplifying this requires server changes anyways.

```json 
{
  MaxEpisodeLength: {
    Type: 'field'
    Name: 'Max Episode Length',
    Description: 'The maximum length to truncate episode names to',
    Size: 'Large',
    Min: 1,
    Max: 250
  }
}
```

PR #1360, opened 2026-06-08
Prepare for pnpm 11

khangp0000:
- Migrate pnpm settings from `package.json` to `pnpm-workspace.yaml`. Both `overrides` and `allowBuilds` are supported in pnpm 10, making this change backward compatible while preparing for pnpm 11 where the `pnpm` field in `package.json` is no longer read.

- `allowBuilds` is required in pnpm 11 because build scripts for dependencies are no longer allowed by default — without it, `pnpm install` will fail with `ERR_PNPM_IGNORED_BUILDS`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,11 +94,6 @@
     "typescript-eslint": "8.60.1",
     "vite": "^8.0.16"
   },
-  "pnpm": {
-    "overrides": {
-      "dompurify": "^3.4.7"
-    }
-  },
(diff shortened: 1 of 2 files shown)


khangp0000:
Alternatively we can use `strictDepBuilds: false` but we can theoretically use `allowBuilds` with version to pin the version of build script dependency to avoid supply chain attack. You decide.

PR #1361, opened 2026-06-08
Add settings modal for plugin update preferences

krbrs:
Surfaces the `Plugins.Updates` server settings (auto-sync repositories, auto-upgrade plugins, update frequency, inactive version retention) through a new modal, opened via a cogwheel button on the Plugin Management page — mirroring the existing dashboard settings modal pattern.

Changes:
- New `PluginUpdateSettingsModal` with local draft state, save/cancel, and its own patch mutation (consistent with `DashboardSettingsModal`)
- Add `PluginUpdatesSettingsType` and `Plugins.Updates` initial defaults (verified against the swagger schema)
- Extend `convertTimeSpanToMs` in `core/util.ts` to handle the optional day-prefix present in longer TimeSpan values (e.g. `"30.00:00:00"`), and add `convertMsToTimeSpan` for the reverse conversion — both used to translate between milliseconds and the API's `"[d.]hh:mm:ss"` TimeSpan format
- Wire the cogwheel button + modal into `PluginManagementSettings`

### DIFF
--- a/src/components/Settings/PluginManagement/Dialogs/PluginUpdateSettingsModal.tsx
+++ b/src/components/Settings/PluginManagement/Dialogs/PluginUpdateSettingsModal.tsx
@@ -1,0 +1,157 @@
+import React, { useEffect, useState } from 'react';
+import { produce } from 'immer';
+import { isEqual, toNumber } from 'lodash';
+
+import Button from '@/components/Input/Button';
+import Checkbox from '@/components/Input/Checkbox';
+import SelectSmall from '@/components/Input/SelectSmall';
+import ModalPanel from '@/components/Panels/ModalPanel';
+import UpdateFrequencyValues from '@/components/Settings/UpdateFrequencyValues';
+import { useVersionQuery } from '@/core/react-query/init/queries';
+import { usePatchSettingsMutation } from '@/core/react-query/settings/mutations';
+import { useSettingsQuery } from '@/core/react-query/settings/queries';
+import { convertMsToTimeSpan, convertTimeSpanToMs, dayjs } from '@/core/util';
+
+import type { PluginUpdatesSettingsType, SettingsUpdateFrequencyType } from '@/core/types/api/settings';
+
+type Props = {
+  onClose: () => void;
+  show: boolean;
+};
+
+const PluginUpdateSettingsModal = ({ onClose, show }: Props) => {
+  const settings = useSettingsQuery().data;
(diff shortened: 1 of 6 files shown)


src/components/Settings/PluginManagement/Dialogs/PluginUpdateSettingsModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  const handleSave = () => {
+    patchSettings(
+      produce(settings, (draft) => {
+        draft.Plugins.Updates = updatesSettings;
+      }),
+      { onSuccess: () => onClose() },
+    );
+  };
```
harshithmohan, line 60: Store the result of produce in a var and then use it in patchSettings for readability

src/components/Settings/PluginManagement/Dialogs/PluginUpdateSettingsModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  <UpdateFrequencyValues min24Hours />
+                </SelectSmall>
+              </div>
+            </div>
+
+            <div className="border-b border-panel-border" />
+          </>
+        )}
+
+        <div className="flex flex-col gap-y-6">
```
harshithmohan: The parent div already has gap-y-6, is this div needed?

src/components/Settings/PluginManagement/Dialogs/PluginUpdateSettingsModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const versionQuery = useVersionQuery();
+  const { isPending: isSavePending, mutate: patchSettings } = usePatchSettingsMutation();
+
+  const isDevChannel = versionQuery.data?.Server.ReleaseChannel !== 'Stable';
+
+  const [updatesSettings, setUpdatesSettings] = useState(settings.Plugins.Updates);
+
+  useEffect(() => {
+    setUpdatesSettings(settings.Plugins.Updates);
+  }, [settings.Plugins.Updates]);
```
harshithmohan: This might not be required since settings would've already loaded by the time we come here, so the initialized value in above line would be correct

harshithmohan (changes requested):
Put the cog inline with the header, subheader can take full width.
It should also have no background, so remove buttonType and make it blue

PR #1382, opened 2026-06-19
Release management

da3dsoul:


hidden4003:
🚨 Critical Issues & Security (Blockers)
1. lastRowIndex.current stale-index crash on filter/search change

MultipleReleasesSeriesList.tsx:185-188

When the user applies a filter (e.g., "Only Finished Series" or a search query), the series array is replaced with a shorter list, but lastRowIndex.current keeps its old value. If a user clicks row 10, then filters down to 5 results, then shift-clicks, series[10] is undefined and the .SeriesID access throws.


// Bug: series may be re-indexed but lastRowIndex is never reset
const targetState = rowSelection?.[series[lastRowIndex.current].SeriesID] ?? true;
Fix: reset lastRowIndex.current = undefined in a useEffect that depends on params (or the series array identity).

🏗️ Architecture & Pattern Checks
2. useMemo / useCallback throughout — CLAUDE.md violation

The project runs the React Compiler, which makes manual memoization redundant and potentially harmful. CLAUDE.md is explicit: "do not use useMemo, useCallback, or React.memo unless absolutely required (e.g., for a library boundary or a measured performance issue)." The new files collectively have 42 uses. None appear to be library-boundary cases. All should be removed:

CandidatesTab.tsx — 7 uses
ManageVariationsModal.tsx, MultipleReleasesDetailModal.tsx, MultipleReleasesPage.tsx, MultipleReleasesSeriesDetailPage.tsx, MixAndMatchTab.tsx, MultipleReleasesSeriesList.tsx
Does this represent an intentional update to the project's React patterns (e.g., an interim state before React Compiler is verified to handle these cases), or should these be removed to align with CLAUDE.md?

3. Dead code: MultipleReleasesDetailModal.tsx is never imported

MultipleReleasesDetailModal.tsx (193 lines) has no importers — only its own file. The page-based approach via MultipleReleasesSeriesDetailPage superseded it. It should be deleted, or if it's intentionally kept for future use, at minimum the AGENTS.md / spec should note why.

4. Duplicated logic between CandidatesTab.tsx and MultipleReleasesDetailModal.tsx

Both files independently define the same three pure helpers:


const computeEffectivePrimary = ...  // identical in both
const buildEpisodeSet = ...          // identical in both
const isSubsetOf = ...               // identical in both
If MultipleReleasesDetailModal is kept, these should be extracted to a shared module (e.g., src/core/utilities/releaseManagementHelpers.ts). If the modal is deleted, this is moot.

5. SignalR handler — DRY regression in eventHandlers.ts

src/core/signalr/eventHandlers.ts: The old code used a clean fallthrough. The new code breaks FileDetected/FileHashed out of the fallthrough but duplicates the three invalidate* calls across three separate cases instead of grouping them. If a new invalidation target is added in the future, all three cases must be updated manually. Consider a shared helper or restoring the fallthrough structure:


// Before (elegant): FileDetected, FileHashed fell through to FileMatched
// After (duplicated across 3 cases): same 3 invalidations copied in each
💡 Suggestions & Refactoring
6. perFileChecks in CandidatesTab is collected but never sent to the server

CandidatesTab.tsx:46 — perFileChecks is maintained via useImmer and passed down to each CandidateCard, so users can check/uncheck individual files. However, when "Preview Deletion for this Series" is clicked, the state is never passed to onPreviewSingle — the mutation just uses the default server logic. The checkboxes render correctly but their values are silently discarded. Either wire them into the preview call or remove the UI to avoid misleading the user.

7. AllowDeletion field has no UI control and stays false

src/core/types/api/settings.ts:328 adds AllowDeletion: boolean to ReleaseComparisonPreferencesType, and src/core/react-query/settings/helpers.ts:406 initializes it to false. But ReleaseManagementSettings.tsx exposes no checkbox for it. If the server requires AllowDeletion: true to actually execute deletions, the feature will silently fail for all users. Verify whether the server uses this field, and if so, add a UI control.

8. eslint-disable comment in MultipleReleasesPreviewModal.tsx is stale

MultipleReleasesPreviewModal.tsx:989-990:


// eslint-disable-next-line react-hooks/exhaustive-deps
}, [open]);
The project uses oxlint, not ESLint. The comment doesn't suppress the oxlint version of this rule. More importantly, the intentional stale-dep pattern here (only re-fetch when the modal opens) is correct behaviorally, but the suppression comment should either use the oxlint equivalent or the useEffect should be refactored to avoid needing suppression at all (e.g., separate the reset logic from the fetch trigger).

9. Query key for useMultipleReleaseSeriesQuery includes the full params object

queries.ts:70:


queryKey: ['release-management', 'multiple-releases', 'series', params],
This is correct and follows the project pattern. However, params.pageSize is always 50 and is included in the key. If pageSize ever varies by context, the cache would not be shared. This is fine as-is, just noting it's intentional.

da3dsoul:
Ready for review again

src/components/Settings/HashingAndReleaseSettings/ReleaseManagementSettings.tsx
```diff
@@ -0,0 +1,241 @@
...
+    path={mdiInformationVariantCircleOutline}
+    size={0.75}
+    className="ml-1 shrink-0 opacity-65"
+    data-tooltip-id="tooltip"
+    data-tooltip-content={tooltip}
+  />
+);
+
+const ReleaseManagementSettings = ({ onChange, preferences }: Props) => {
+  const set = <K extends keyof ReleaseComparisonPreferencesType>(
```
harshithmohan: name that `updateSetting` or something

src/components/Settings/HashingAndReleaseSettings/ReleaseManagementSettings.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+const InfoIcon = ({ tooltip }: { tooltip: string }) => (
+  <Icon
+    path={mdiInformationVariantCircleOutline}
+    size={0.75}
+    className="ml-1 shrink-0 opacity-65"
+    data-tooltip-id="tooltip"
+    data-tooltip-content={tooltip}
+  />
+);
```
harshithmohan: We show the info below the setting itself, look at other settings in the same page. We don't need an info icon

src/components/Settings/HashingAndReleaseSettings/ReleaseManagementSettings.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+      <div className="flex flex-col gap-y-2">
+        <div className="font-semibold">Quality Preferences</div>
+
+        <div className="flex items-center justify-between">
+          <span className="flex items-center">
+            Episode Type Scope
+            <InfoIcon tooltip="Controls how releases covering different episode types (regular vs. specials) are ranked. 'All Together' treats them as one unit; 'Per Episode Type' ranks them independently." />
+          </span>
+          <Select
```
harshithmohan: This should be `SelectSmall`

src/components/Settings/HashingAndReleaseSettings/ReleaseManagementSettings.tsx
```diff
@@ -0,0 +1,241 @@
...
+          label={
+            <span className="flex items-center">
+              Prefer Higher Bit Depth
+              <InfoIcon tooltip="Prefer 10-bit video over 8-bit when ranking candidates. Disable to prefer 8-bit." />
+            </span>
+          }
+        />
+
+        <div className="flex items-center justify-between">
+          <span className="flex items-center text-sm">
```
harshithmohan: Why are these setting labels smaller than normal size?
This and the 4 below it

src/core/react-query/release-management/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type ReleaseDeletionPreviewBody = {
+  includedSeriesIDs?: number[];
+  excludedSeriesIDs?: number[];
+  overrides?: SeriesCandidateOverride[];
+};
+
+type DeleteReleasesBody = {
+  placeIDs: number[];
+};
```
harshithmohan: These should in types.ts

src/core/react-query/release-management/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const useReleaseDeletionPreviewMutation = () =>
+  useMutation({
+    mutationFn: (body: ReleaseDeletionPreviewBody): Promise<ReleaseDeletionPreviewType[]> =>
+      axios.post('ReleaseManagement/MultipleReleases/Preview', body),
+  });
+
+export const useReleaseOverrideMutation = (seriesId: number) =>
+  useMutation({
+    mutationFn: (body: ReleaseOverrideBody): Promise<ReleaseDeletionPreviewType> =>
```
harshithmohan: For consistency, we shouldn't specify return types of the function here, they should be inferred by setting them on `useMutation`, the body type also will be specified with `useMutation` if return type is specified. Check other files for reference

src/core/react-query/release-management/queries.ts
```diff
@@ -62,3 +64,35 @@ export const useReleaseManagementSeriesEpisodes = (
...
+    initialPageParam: 1,
+    getNextPageParam: (lastPage, _, lastPageParam: number) => {
+      if (!params.pageSize || lastPage.Total / params.pageSize <= lastPageParam) return undefined;
+      return lastPageParam + 1;
+    },
+  });
+
+export const useMultipleReleaseSeriesDetailQuery = (seriesId: number, includeVariations = false, enabled = true) =>
+  useQuery<SeriesWithCandidatesType>({
+    queryKey: ['release-management', 'multiple-releases', 'series', seriesId, { includeVariations }],
```
harshithmohan: Why is `includeVariations` inside an object?

src/core/react-query/release-management/queries.ts
```diff
@@ -62,3 +64,35 @@ export const useReleaseManagementSeriesEpisodes = (
...
+    queryFn: () =>
+      axios.get(
+        `ReleaseManagement/MultipleReleases/Series/${seriesId}`,
+        includeVariations ? { params: { includeVariations: true } } : undefined,
```
harshithmohan: The ternary is useless, the value can be either true or false and if we don't pass it, it's set as false anyway on line 88

src/core/signalr/eventHandlers.ts
```diff
@@ -42,6 +47,7 @@ export const handleEvent = (event: string, data?: SeriesUpdateEventType) => {
       invalidateDashboard();
       invalidateFiles();
       invalidateManagedFolders();
+      if (event === 'FileDeleted' || event === 'FileMatched') invalidateReleaseManagement();
```
harshithmohan, line 50: (placeholder comment to discuss removal)

src/core/types/api/release-management.ts
```diff
@@ -0,0 +1,109 @@
...
+export type ReleaseCandidateType = {
+  Name: string;
+  Rank: number;
+  Key: string;
+  HasReleaseInfo: boolean;
+  IsRedundant: boolean;
+  IsPartial?: boolean;
+  DecidingSignal: string | null;
+  WinnerValue: string | null;
+  LoserValue: string | null;
```
harshithmohan: We know they are not optional and can return null but write them as 
`DecidingSignal?: string;`

src/core/utilities/buildEpisodeCoverageString.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const typeDisplayName = (type: string): string => {
+  if (type === 'Episode') return 'Episodes';
+  if (type === 'Special') return 'Specials';
+  if (type === 'Trailer') return 'Trailers';
+  if (type === 'ThemeSong') return 'Theme Songs';
+  if (type === 'Credits') return 'Credits';
+  if (type === 'Parody') return 'Parodies';
+  return type;
+};
```
harshithmohan: That could be an object

harshithmohan (changes requested):

src/core/types/api/settings.ts
```diff
@@ -323,6 +323,20 @@ export type SettingsImportType = {
...
+  SourceOrder: string[];
+  ResolutionOrder: string[];
+  VideoCodecOrder: string[];
+  AudioCodecOrder: string[];
+  SubGroupOrder: string[];
+  PreferHigherBitDepth: boolean;
+  AllowDeletion: boolean;
+  AutoDeleteOnImport: boolean;
+  PerFileDeletionForAiringSeries: boolean;
+  EpisodeTypeScope: 0 | 1;
```
harshithmohan: For `SignalPriority` and `EpisodeTypeScope`
The existing behaviour for any enums is for them to be returned as a string in the API and also take strings in the requests, we do not use magic numbers. So these should ideally use `StringEnumConverter` on the server.

harshithmohan:
image
When we go in to Auto delete mode, the cancel button (I'm assuming that's what it is because it takes me back) still says "auto delete" instead of cancel

harshithmohan:
https://github.com/user-attachments/assets/64ad0576-36cc-4ca4-9765-43464e71199e

If we deselect all in the auto-deletion, the delete button is still enabled. And not only that but we end up with more files to delete than actual list somehow

harshithmohan:
image

It can just say "Preview Deletion" since we are already inside a series.
Also is the same button being at the end of the page as well intentional?

harshithmohan:
https://github.com/user-attachments/assets/d67bd2f0-f636-492b-afb0-0f1d7729094f

Auto fill should work if you click on another group even if you don't clear it first

harshithmohan:
All rows should have alternating background colours, refer to other utils

src/components/Utilities/ReleaseManagement/CandidateCard.tsx
```diff
@@ -0,0 +1,419 @@
...
+          Would be deleted
+        </span>
+      );
+    }
+    return (
+      <span className="rounded-sm bg-panel-background-alt px-1.5 py-0.5 text-xs font-semibold text-panel-text">
+        No auto-delete available
+      </span>
+    );
+  };
```
harshithmohan: We have a `Badge` component to use

harshithmohan: Non-component functions which return components shouldn't be used anyway

src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesList.tsx
```diff
@@ -0,0 +1,248 @@
...
+          onChange={onToggle}
+          label=""
+        />
+      </div>
+
+      <div className="flex min-w-0 grow flex-col gap-0.5">
+        <div className="flex items-center gap-2">
+          <span className="truncate font-semibold">{series.SeriesTitle}</span>
+          {series.IsAiring && (
+            <span className="shrink-0 rounded-sm bg-panel-text-primary px-1.5 py-0.5 text-xs font-semibold text-white">
```
harshithmohan: `text-white` shouldn't be used as it's a tailwind inbuilt colour.

src/components/Utilities/ReleaseManagement/CandidateCard.tsx
```diff
@@ -0,0 +1,419 @@
...
+      return candidate.Version > 1 ? `Best Available (up to v${candidate.Version})` : 'Best Available';
+    }
+    return null;
+  })();
+
+  const redundancyBadge = () => {
+    if (isPrimary) return null;
+    if (candidate.IsRedundant) {
+      return (
+        <span className="rounded-sm bg-panel-text-danger px-1.5 py-0.5 text-xs font-semibold text-white">
```
harshithmohan: text-white shouldn't be used as it's a tailwind inbuilt colour.

hidden4003:
## PR Review Follow-up

Great work addressing the review — the follow-up commits resolved most findings. Here's the full status:

### ✅ Resolved

| # | Finding |
|---|---------|
| 1 | `lastRowIndex.current` stale-index crash — `useEffect` resetting the ref on filter/search changes added |
| 2 | `useMemo`/`useCallback` throughout — fully removed from all new components |
| 3 | `MultipleReleasesDetailModal` dead code — deleted |
| 4 | Duplicated helpers (`buildEpisodeSet`, etc.) — `buildEpisodeSet` moved to shared utility; the modal holding the other copies was deleted |
| 5 | SignalR DRY regression — fallthrough restored, `invalidateReleaseManagement()` called conditionally inside the shared case |
| 6 | `perFileChecks` collected but never sent to the server — fully removed from `CandidatesTab` |
| 7 | Stale `eslint-disable` comment — updated to `// oxlint-disable-next-line react-hooks/exhaustive-deps` |
| 8 | Tab bar used template literals instead of `cx()` — now uses `cx()` |
| 9 | `EpisodeTypeScope: number` not narrowed — changed to `0 | 1`, matching the project's existing pattern |

### ⚠️ Still Open

**1. `AllowDeletion` has no UI control and is hardcoded to `false`**

`AllowDeletion: boolean` is defined in `ReleaseComparisonPreferencesType` and initialized to `false` in `initialSettings`, but `ReleaseManagementSettings.tsx` has no toggle for it. If the server gates deletion on this flag, deletion will silently fail for all users. Is this intentional (managed server-side only) or a missing control?

**2. Inconsistent `rel` attribute on AniDB links**

- `MultipleReleasesSeriesList.tsx`: `rel="noreferrer noopener"`
- `MultipleReleasesSeriesDetailPage.tsx`: `rel="noreferrer"`

Since `noreferrer` implies `noopener`, the shorter form in the detail page is correct. Please update the list component to match.

**3. Double-space separator in `buildEpisodeCoverageString`**

```ts
.join('  ')  // two spaces between episode type groups
```
This looks like a typo — a ' · ' or ', ' would be more readable and clearly intentional.

hidden4003:
## Release Management PR — Code Review Findings

---

### 🔴 Correctness Bugs

---

#### 1. Infinite-scroll series silently excluded from deletion preview despite appearing checked

**File:** [MultipleReleasesPage.tsx](src/components/Utilities/ReleaseManagement/MultipleReleasesPage.tsx#L84)

`handleToggleAutoDelete` pre-fills `rowSelection` with only the IDs present at that instant. As the user scrolls and more series load, `onSeriesIdsChange` updates `allSeriesIds` but never adds new IDs to `rowSelection`. `SeriesRow` renders those new series as checked (`rowSelection?.[id] ?? true` defaults to `true`), but:

```ts
includedForPreview = allSeriesIds.filter(id => rowSelection[id])
```

…has no `?? true` fallback — `undefined` is falsy — so those series are excluded. The user confirms what looks like a full-list deletion, but a silent subset is actually queued.

---

#### 2. Shift-click range selection toggles in the wrong direction for late-loaded series

**File:** [MultipleReleasesSeriesList.tsx](src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesList.tsx#L179)

When the shift-click anchor is a series loaded after auto-delete mode was activated (no explicit `rowSelection` entry):

```ts
targetState = rowSelection?.[series[lastRowIndex.current].SeriesID] ?? true
//                                                                    ^^^^^^^^ always true for untracked rows
```

Every row in the range is set to `selected = true` even when the user's intent was to deselect. Wrong series end up queued for deletion.

---

#### 3. "Clear All" is a no-op when every episode has exactly one file option

**File:** [MixAndMatchTab.tsx](src/components/Utilities/ReleaseManagement/MixAndMatchTab.tsx#L208)

```ts
const handleClearAll = () => {
  setSelections(autoSelectSingleOptions(allEpisodes));
};
```

`autoSelectSingleOptions` pre-selects every episode where `options.length === 1`. For a series where all episodes have exactly one file, the returned `Map` is identical to the current state — clicking **Clear All** does nothing. There is no disabled state or tooltip to inform the user.

---

#### 4. R keybinding resets the wrong React Query key on the Multiple Releases tab

**File:** [ReleaseManagement.tsx](src/pages/utilities/ReleaseManagement.tsx#L91)

`useHotkeys('r', handleRefresh)` is registered unconditionally before the early return for `type === 'MultipleReleases'`. `handleRefresh` resets `['release-management', 'series']`, but `MultipleReleasesPage` uses `['release-management', 'multiple-releases']`. `MenuButton`'s `keybinding="R"` prop is purely cosmetic (no actual hotkey registration). Pressing R resets an unused query while the visible list never refreshes.

---

### 🟡 Consistency Bugs

---

#### 5. "Airing" badge has two different visual treatments within the same PR

| Location | Classes |
|---|---|
| [MultipleReleasesSeriesList.tsx:93](src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesList.tsx#L93) | `bg-panel-text-primary text-white` (solid) |
| [MultipleReleasesSeriesDetailPage.tsx:118](src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesDetailPage.tsx#L118) | `bg-panel-text-primary/20 text-panel-text-primary` (translucent) |

The same "this series is airing" concept diverges immediately at introduction with no shared constant or component to keep them in sync.

---

### 🔵 Reuse / Duplication

---

#### 6. Seven-plus inline badge spans fork the Badge component with a different shape

**File:** [CandidateCard.tsx](src/components/Utilities/ReleaseManagement/CandidateCard.tsx#L71)

`src/components/Badge.tsx` is the shared badge primitive (`rounded-lg px-2.5 py-1 text-xs font-semibold`). `CandidateCard` alone has 7+ one-off inline `<span>`s using `rounded-sm px-1.5 py-0.5 text-xs font-semibold` (redundancy status, MIXED, strategy label, version delta, PARTIAL, file anomaly tags), with more in `MultipleReleasesSeriesDetailPage` and `MultipleReleasesSeriesList`. Any future change to badge sizing or radius must be hunted across all of them individually rather than updated in one place.

---

#### 7. Inline tab bar duplicates TabPills and is missing its `disabled={active}` guard

**File:** [MultipleReleasesSeriesDetailPage.tsx:147](src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesDetailPage.tsx#L147)

`src/components/TabPills.tsx` already handles active state, overflow, and divided styling. The inline Candidates / Mix & Match switcher hand-rolls the same two-button + `cx`-conditional pattern but omits `TabPills`'s `disabled={active}` check — clicking the currently active tab fires a redundant `setSearchParams` update on every click.

---

#### 8. Filename extraction copy-pasted verbatim into three files

```ts
file.AbsolutePath?.split(/[/\\]/).pop() ?? `Place ${file.PlaceID}`
```

Appears identically in:
- [CandidateCard.tsx:337](src/components/Utilities/ReleaseManagement/CandidateCard.tsx#L337)
- [MultipleReleasesPreviewModal.tsx:78](src/components/Utilities/ReleaseManagement/MultipleReleasesPreviewModal.tsx#L78)
- [MixAndMatchTab.tsx:345](src/components/Utilities/ReleaseManagement/MixAndMatchTab.tsx#L345)

Three places to update if the fallback label or path-separator regex ever changes.

---

### ⚪ CLAUDE.md Violations

---

#### 9. `||` used on non-boolean operands — should be `??`

**File:** [CandidateCard.tsx](src/components/Utilities/ReleaseManagement/CandidateCard.tsx)

> *CLAUDE.md: "Use `??` instead of `||`, except for boolean values where `||` is acceptable."*

| Line | Expression | Type |
|---|---|---|
| [229](src/components/Utilities/ReleaseManagement/CandidateCard.tsx#L229) | `candidate.Version \|\| 1` | `number` — also masks a legitimate `0` |
| [280](src/components/Utilities/ReleaseManagement/CandidateCard.tsx#L280) | `coverageString \|\| 'None'` | `string` |

---

#### 10. `Map`/`Set` state uses `useState` + manual copy-mutate instead of `useImmer`

**File:** [MixAndMatchTab.tsx:170](src/components/Utilities/ReleaseManagement/MixAndMatchTab.tsx#L170)

> *CLAUDE.md: "Immutable state → `immer` or `useImmer`."*

`selections` (`Map<string, number>`) and `expandedEpisodes` (`Set<string>`) both use `useState` with copy-mutate patterns (`new Map(prev).set(...)`, `new Set(prev)`). `useImmer` draft mutations are the documented pattern and would be less verbose, especially in `handleAutofill` which iterates all episodes and conditionally inserts into a copied `Map`.

da3dsoul:
> image
> It can just say "Preview Deletion" since we are already inside a series. Also is the same button being at the end of the page as well intentional?

The button being at the bottom was intentional. The first time I tested it was on a series with over 50 episodes, and scrolling to the top seemed annoying

src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesList.tsx
da3dsoul: What should be used

src/core/utilities/buildEpisodeCoverageString.ts
da3dsoul: I don't know what that means in this context

harshithmohan:
```suggestion
export const typeDisplayNameMap = {
  Episode: 'Episodes',
  Special: 'Specials',
  Trailer: 'Trailer',
  ThemeSong: 'Theme Songs',
  Credits: 'Credits',
  Parody: 'Parodies'
};
```

src/components/Utilities/ReleaseManagement/MultipleReleasesSeriesList.tsx
harshithmohan: Either the default color or the one you want to from our theme vars, but usually the default, i.e., remove the class and don't override the color

hidden4003:
### 📋 PR Review Summary

This update (commit `5c1fde1e`) adds richer per-file quality data to `CandidateCard` and `MixAndMatchTab`, shows directory paths in the preview modal, and refactors anomaly badge rendering. All prior round findings remain resolved. Two new correctness bugs were introduced in this update — one moderate, one minor visual.

---

### 🚨 Critical Issues & Security (Blockers)

None detected.

---

### 🏗️ Architecture & Pattern Checks

**Previous findings — confirmed still resolved:**

All findings from prior review rounds remain addressed. No regressions in useMemo/useCallback, SignalR DRY, stale-index crashes, rowSelection model, useImmer usage, or tab guard.

---

### 🔴 Correctness Bugs (New)

**1. Series-detail deletion does not reset the query cache before navigating back**

`mutations.ts` removed `resetQueries(['release-management'])` from `useReleaseExecuteMutation.onSuccess` and moved invalidation to callers. `MultipleReleasesPage.tsx` correctly passes:

```tsx
onSuccess={() => resetQueries(['release-management'])}
```

But `MultipleReleasesSeriesDetailPage.tsx` only passes:

```tsx
onSuccess={() => navigate(`/webui/utilities/release-management/MultipleReleases...`)}
```

After deletion, the user lands back on the series list with cached `staleTime: Infinity` data — the just-deleted/modified series still appears until a SignalR `FileDeleted` event arrives (debounced 2 s). If SignalR is slow or disconnected, the stale list persists indefinitely.

**Fix:** Add a cache reset to the series detail's `onSuccess`:

```tsx
onSuccess={() => {
  resetQueries(['release-management', 'multiple-releases']);
  navigate(`/webui/utilities/release-management/MultipleReleases${listSearch ? `?${listSearch}` : ''}`);
}}
```

---

**2. `signalLabels` keys in `CandidateCard.tsx` don't match `SignalType` / what the server sends as `DecidingSignal`**

`CandidateCard.tsx` (line 29) defines:

```ts
const signalLabels: Record<string, string> = {
  AudioStreamCount: 'Audio Streams',   // server sends 'AudioStreams'
  SubtitleStreamCount: 'Subtitle Streams', // server sends 'SubtitleStreams'
  Chapters: 'Chapters',               // server sends 'Chaptered'
  IsCorrupted: 'Corruption',          // server sends 'Corrupted'
  IsCensored: 'Censorship',           // server sends 'Censored'
  GroupHomogeneity: 'Group Consistency', // not in SignalType at all
  // Missing: 'Creditless'
  ...
};
```

Compare to `SignalType` in `settings.ts` and `SIGNAL_NAMES` in `ReleaseManagementSettings.tsx`, which use `AudioStreams`, `SubtitleStreams`, `Chaptered`, `Corrupted`, `Censored`, `Creditless`. The lookup `signalLabels[candidate.DecidingSignal]` falls back to `?? candidate.DecidingSignal` for misses — so these signals display as raw strings (`'AudioStreams'`, `'Chaptered'`, etc.) rather than humanized labels.

**Fix:** Align the keys with what the server sends:

```ts
const signalLabels: Record<string, string> = {
  Source: 'Source',
  Resolution: 'Resolution',
  VideoCodec: 'Video Codec',
  BitDepth: 'Bit Depth',
  AudioCodec: 'Audio Codec',
  AudioLanguage: 'Audio Language',
  AudioStreams: 'Audio Streams',
  SubtitleLanguage: 'Subtitle Language',
  SubtitleStreams: 'Subtitle Streams',
  Version: 'Version',
  Chaptered: 'Chaptered',
  Censored: 'Censorship',
  Creditless: 'Creditless',
  Corrupted: 'Corruption',
  SubGroup: 'Sub Group',
};
```

---

### 💡 Suggestions & Refactoring

**3. R hotkey guard watches the wrong fetching state on the Multiple Releases page** *(carried from last review)*

`ReleaseManagement.tsx` guards `handleRefresh` with:
```ts
const isSeriesQueryFetching = useIsFetching({ queryKey: ['release-management', 'series'] }) > 0;
```

But when `type === 'MultipleReleases'` that counter is always 0, so R can trigger a reset mid-fetch. Fix by branching the query key:

```ts
const isSeriesQueryFetching = useIsFetching({
  queryKey: type === 'MultipleReleases'
    ? ['release-management', 'multiple-releases']
    : ['release-management', 'series'],
}) > 0;
```

---

### 🔍 Nitpicks

**4. Empty badge container renders for non-chaptered files**

`CandidateCard.tsx` (line 398): the outer condition for the anomaly badge block is:

```tsx
{(file.IsChaptered != null || file.IsCensored || file.IsCreditless || file.IsCorrupted) && (
  <div className="flex flex-wrap gap-1 pt-0.5">
    {file.IsChaptered != null && (file.IsChaptered || isChapteredAnomaly) && (...)}
    {file.IsCensored && (...)}
    {file.IsCreditless && (...)}
    {file.IsCorrupted && (...)}
  </div>
)}
```

For a recognized, non-chaptered, non-censored, non-corrupted file: `IsChaptered = false`, so `file.IsChaptered != null` is `true` — the container renders. But all inner conditions are false, so you get an empty `<div>` with `pt-0.5` padding. Since most recognized files are non-chaptered, this affects the majority of file rows. Fix the outer guard:

```tsx
{((file.IsChaptered != null && (file.IsChaptered || isChapteredAnomaly))
  || file.IsCensored
  || file.IsCreditless
  || file.IsCorrupted) && (
```

**5. Bug 6 (inline `<span>` vs `<Badge>`)** *(from prior review)* — **Closing this finding.** The file-level anomaly tags have context-aware conditional styling (`cx()` selecting between warning/neutral styles) that the shared `<Badge>` component isn't designed to handle. The differentiation is now intentional.

---

**Bottom line:** Two things need fixing before this merges: the missing cache reset in the series-detail delete flow (#1), and the `signalLabels` key mismatch (#2) which silently leaves several signals un-humanized. The nitpick (#4) is a real but invisible layout issue worth a one-liner fix.

da3dsoul:
I think it's gaslighting a bunch of that. I'll verify

hidden4003:
### 📋 PR Review Summary — Round 5

Reviewed commits `52c1dd9` (Address PR Review. Unified Signals in Server), `d018ccf` (Cleanup and let claude review), `84ace43` (Fix lint issue). All round 4 findings are resolved. No new blockers — two behavior changes worth confirming and one nitpick remains open.

---

### 🚨 Critical Issues & Security (Blockers)

None detected.

---

### 🏗️ Architecture & Pattern Checks

**All previous findings — confirmed still resolved:**

- Auto Delete button: two-button "Cancel"/"Auto Delete" pattern restored — ✅
- R hotkey: `isSeriesQueryFetching` and `resetQueries` now branch on `type` — ✅
- Per-file stream summary and language rows restored in `CandidateCard` — ✅
- `signalLabels` unified in `buildEpisodeCoverageString.ts` as `Record<SignalType, string>`, shared by both `CandidateCard` and `ReleaseManagementSettings` — cleaner and correct — ✅
- `EpisodeTypeScope` migrated from `0 | 1` to `EpisodeTypeScopeType = 'AllTogether' | 'PerEpisodeType'` — ✅
- `AudioLanguageOrder`, `SubtitleLanguageOrder` fields added to settings type and UI — ✅
- `onlyFinishedSeries` now flows through to the preview mutation from the list page — ✅
- Body types extracted from `mutations.ts` to `types.ts` — ✅
- `MixAndMatchTab` selection state uses `useImmer` — ✅
- `SeriesList` resets selected series on `type` change rather than on data change — ✅

---

### 🔴 Must Fix

None.

---

### 💡 Suggestions

**1. `handleAutofill` now overwrites existing manual selections — confirm intent**

`MixAndMatchTab.tsx` — `handleAutofill` previously guarded with `!next.has(episode.key)` so it only filled unassigned slots. It now sets every episode that has an option in the target group, overwriting prior manual picks:

```tsx
// before: only fill unassigned
if (!next.has(episode.key)) { next.set(...); }

// now: overwrite all matching
draft.set(episode.key, best.placeID);
```

If the intended UX is "apply this group to all episodes" (similar to a 'Select All from Group X' action), the new behavior is correct. If the intent was to only fill gaps, restore the guard. The label says "Autofill [Group]" which could go either way — worth a comment or tooltip clarification.

**2. `handleClearAll` now fully clears rather than resetting to auto-selections**

Same file — `handleClearAll` used to call `autoSelectSingleOptions(allEpisodes)`, which re-applied the initial auto-selections (single-option episodes got auto-picked). It now calls `draft.clear()`, leaving all rows unassigned.

This is more consistent with "Clear All" semantics but means episodes with only one option also lose their auto-selection. The user must then autofill or manually select before the Delete button becomes active. Confirm this is the intended experience.

---

### 🔍 Nitpicks

**3. Redundant optional chaining in `CandidatesTab.tsx`** *(carried from round 4)*

`computeEffectivePrimary` has return type `ReleaseCandidateType` (non-nullable). Lines 49 and 78 correctly use `effectivePrimary.Episodes` and `effectivePrimary.Key` without optional chaining. Lines 115 and 123 still use `effectivePrimary?.Key` and `effectivePrimary?.Episodes ?? []`. Remove the `?.` and `?? []` on those two lines for consistency.

**4. `search` param sends empty string when no filter is active**

`MultipleReleasesSeriesList.tsx` reverted from `search: search ?? undefined` back to `search: search`. When the URL has no `?search=` param, `searchQuery` resolves to `''`, which is now sent as `?search=` in the API request. The server likely treats this as no-op, but it's unnecessary noise on every request. Consider `search: search || undefined`.

---

### Notable improvements in this round (not findings)

- `SignalPriority` type changed from `number[]` to `SignalType[]` — matches "Unified Signals in Server" commit. Requires server to now send string-based signal keys rather than numeric indices. Assuming this is coordinated with the server PR.
- `buildOptionSummary` in `MixAndMatchTab` now includes video codec, bit depth, and audio codec — consistent with `buildFileStreamSummary` in `CandidateCard`.
- Language info added per-option in Mix & Match tab — consistent with `CandidateCard`.
- Alternating row colors added to Mix & Match episode accordion — visual polish.

---

**Bottom line:** This is merge-ready. No blockers remain. The two behavior changes (#1 autofill, #2 clear-all) are worth a comment confirmation from da3dsoul but don't block merge. The optional chaining nitpick (#3) is a one-liner cleanup. The empty-string search param (#4) is harmless.

hidden4003:
> I think it's gaslighting a bunch of that. I'll verify

I deleted that one it did not pull all the changes. Says all good now.

PR #1383, opened 2026-06-20
Sort by Relevance

da3dsoul:
Take advantage of server changes. It will fall back on regular name sorting if it ends up searching by ID. It wouldn't matter, since that would return one result, anyway.

### DIFF
--- a/src/pages/collection/Collection.tsx
+++ b/src/pages/collection/Collection.tsx
@@ -55,16 +55,16 @@ const getFilter = (
           },
           Parameter: query,
         },
-      };
-    }
+  let finalSortingCriteria = sortingCriteria;
+const cleanFilterConditions = filterConditions.filter(condition => !!condition);
+if (query) {
+let searchCondition: FilterCondition = {
+@@ -62,6 +63,10 @@ const getFilter = (
+} else {
+finalCondition = buildFilter([searchCondition]);
+}
 
-    if (cleanFilterConditions.length > 0) {
-      finalCondition = buildFilter([searchCondition, ...cleanFilterConditions]);
-    } else {
-      finalCondition = buildFilter([searchCondition]);
-    }
-
-    finalSortingCriteria = sortingCriteria === undefined
+    finalSortingCriteria = !sortingCriteria
       ? { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false }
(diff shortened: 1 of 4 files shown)


src/pages/collection/Collection.tsx
```diff
@@ -71,7 +71,7 @@ const getFilter = (
       ? {
         ApplyAtSeriesLevel: true,
         Expression: finalCondition,
-        Sorting: sortingCriteria ?? { Type: 'Name', IsInverted: false },
+        Sorting: sortingCriteria ?? { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false },
```
harshithmohan: The webui shouldn't assume the server will fallback to name sorting if query is not provided, it should only use fuzzy *if* query is non-empty

harshithmohan (changes requested):

src/pages/collection/Collection.tsx
da3dsoul: Shouldn't it not filter at all if there's no query?

harshithmohan: If there's no query and another filter is applied*

da3dsoul: Realistically, that's just the server handling a failure mode gracefully. The server has always fallen back on sorting by name when none is provided. I can make it more intentional on the webui side.

src/pages/collection/Collection.tsx
```diff
@@ -62,6 +63,10 @@ const getFilter = (
     } else {
       finalCondition = buildFilter([searchCondition]);
     }
+
+    finalSortingCriteria = sortingCriteria === undefined
+      ? { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false }
+      : { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false, Next: sortingCriteria };
```
harshithmohan:
```suggestion
  let finalSortingCriteria = sortingCriteria;
const cleanFilterConditions = filterConditions.filter(condition => !!condition);
if (query) {
let searchCondition: FilterCondition = {
@@ -62,6 +63,10 @@ const getFilter = (
} else {
finalCondition = buildFilter([searchCondition]);
}

    finalSortingCriteria = !sortingCriteria
      ? { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false }
      : { Type: 'FuzzyNameRelevance', Parameter: query, IsInverted: false, Next: sortingCriteria };
```

PR #1386, opened 2026-06-27
Fix live filter rebuild for tag and boolean criteria

krbrs:
## Summary

This fixes the live filter regression introduced by https://github.com/ShokoAnime/Shoko-WebUI/commit/b50639d05cb54da52d9fe52293715fcbf288dab9, where updating tag or boolean criteria no longer reliably rebuilt the active filter.

The root cause was the shared sidebar filter builder reading Redux state internally through `store.getState()`. That made the dependency invisible to React, so live updates were not tracked correctly.

## What changed

- Made `buildSidebarFilter()` pure by passing the current sidebar filter slices in explicitly
- Updated `FilterSidebar` to provide:
  - `filterConditions`
  - `filterTags`
  - `filterValues`
  - `filterMatch`
- Removed the render-time workaround that only existed to force React to notice `filterConditions`

## Verification

Checked the live UI against the running app and confirmed:
- boolean criteria still rebuild the live preview
- tag criteria save and render correctly
- multivalue criteria save and render correctly

## Notes

This is a root-cause fix for the regression from https://github.com/ShokoAnime/Shoko-WebUI/commit/b50639d05cb54da52d9fe52293715fcbf288dab9, not a tag-only workaround.

### DIFF
--- a/src/components/Collection/Filter/FilterSidebar.tsx
+++ b/src/components/Collection/Filter/FilterSidebar.tsx
@@ -92,10 +92,6 @@ const FilterSidebar = () => {
   const isFilterValid = keys(selectedCriteria).length > 0
     && keys(selectedCriteria).length === keys(activeCriteriaWithValues).length;
 
-  // buildSidebarFilter reads filterConditions internally via store.getState(),
-  // which the React Compiler can't trace. Explicitly reading selectedConditions
-  // here teaches the compiler this expression depends on it. Also guards against
-  // building a filter when no conditions are set.
   const finalFilterExpression = isFilterValid
     ? buildSidebarFilter(
       Object.values(selectedCriteria),
--- a/src/components/Collection/Filter/FilterSidebar.tsx
+++ b/src/components/Collection/Filter/FilterSidebar.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { useParams } from 'react-router';
 import { mdiContentSaveOutline, mdiFilterPlusOutline } from '@mdi/js';
-import { keys, map, values } from 'lodash';
+import { keys, map } from 'lodash';
 
 import AddCriteriaModal from '@/components/Collection/Filter/AddCriteriaModal';
 import DefaultCriteria from '@/components/Collection/Filter/DefaultCriteria';
@@ -84,6 +84,9 @@ const FilterSidebar = () => {
   const dispatch = useDispatch();
   const selectedCriteria = useSelector(state => state.collection.filterCriteria);
   const selectedConditions = useSelector(state => state.collection.filterConditions);
+  const selectedMatch = useSelector(state => state.collection.filterMatch);
+  const selectedTags = useSelector(state => state.collection.filterTags);
+  const selectedValues = useSelector(state => state.collection.filterValues);
   const activeCriteriaWithValues = useSelector(selectActiveCriteriaWithValues);
 
   const isFilterValid = keys(selectedCriteria).length > 0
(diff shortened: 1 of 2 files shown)


src/components/Collection/Filter/FilterSidebar.tsx
hidden4003: remove this comment

hidden4003 (changes requested):
# PR #1386 Review — Fix live filter rebuild for tag and boolean criteria

## PR Review Summary

This PR fixes a live filter regression where updating tag or boolean criteria didn't reliably rebuild the active filter. The root cause was `buildSidebarFilter()` reading Redux state via `store.getState()` — invisible to the React Compiler — so live updates weren't tracked. The fix correctly makes the function pure by injecting the required state slices explicitly from the component. The approach is architecturally sound and complete.

---

## 🚨 Critical Issues & Security (Blockers)

None detected.

---

## 🏗️ Architecture & Pattern Checks

**State injection pattern is correct and complete.** `store.getState()` inside a utility function called during render is a known React anti-pattern with the compiler because it introduces an untracked side-channel. Passing state explicitly via `useSelector` and threading it in is the right fix. The `store` import in `filter.ts:7` is still needed for `addFilterCriteriaToStore` on line 127, so no orphan import.

**`SidebarFilterState` type** is a small duplication of the collection slice's shape, but it's appropriate — it documents exactly what the function needs rather than leaking the full Redux state type into a utility. No concern.

**`Object.values()` vs lodash `values()`** — the PR switched from `values(selectedCriteria)` (lodash) to `Object.values(selectedCriteria)` (native) on `FilterSidebar.tsx:101` and removed `values` from the lodash import. This is fine; both produce the same result and native is lighter.

---

## 💡 Suggestions & Refactoring

### 1. Stale comment block needs removal (`FilterSidebar.tsx:95-98`)

The comment still describes the old bug, not the current code:

```tsx
// Before (stale — describes the workaround that no longer exists):
// buildSidebarFilter reads filterConditions internally via store.getState(),
// which the React Compiler can't trace. Explicitly reading selectedConditions
// here teaches the compiler this expression depends on it. Also guards against
// building a filter when no conditions are set.
const finalFilterExpression = isFilterValid
  ? buildSidebarFilter(...)
  : undefined;
```

```tsx
// After (remove the comment entirely — the code is self-explanatory):
const finalFilterExpression = isFilterValid
  ? buildSidebarFilter(
      Object.values(selectedCriteria),
      { filterConditions: selectedConditions, filterMatch: selectedMatch, ... },
    )
  : undefined;
```

The function is now pure and the state injection makes the dependencies self-evident to both the reader and the compiler. The comment actively misleads future readers.

### 2. Pre-existing: recursive multivalue functions drop `operator` after the first node (`filter.ts:54`, `filter.ts:69`)

This bug predates this PR, but it's worth flagging since the function is now being more carefully reasoned about:

```ts
// buildSidebarFilterConditionMultivalue — operator only applied at first level
Right: buildSidebarFilterConditionMultivalue(conditionValues.slice(1), type),
//                                                                          ↑ operator missing
```

For a 3-value `Or` filter, only the top node gets `Type: 'Or'`; deeper nodes fall back to the `'And'` default. Not introduced here, but should be a follow-up fix.

---

## 🔍 Nitpicks

- `FilterSidebar.tsx:86` — `selectedConditions` is now used both for passing to `buildSidebarFilter` and implicitly still listed in the `useEffect` dependency array. That's correct and fine, but the variable name `selectedConditions` vs the type field `filterConditions` is a mild naming inconsistency across the two files. Not a blocker.

---

## Verdict

**Approve with one non-blocking fix.** The stale comment on lines 95-98 of `FilterSidebar.tsx` should be removed before merge — it actively describes behavior that no longer exists and will mislead future readers. Everything else is clean.

PR #1395, opened 2026-07-17
Improve series poster

pokyunn:
- Add shortcut to "upcoming anime" to directly add them to collection
Screenshot_20260716_200303
- Change indicator icon to distinguish watched/unwatched/without files
Screenshot_20260716_200541


src/components/Collection/PosterViewItem.tsx
```diff
@@ -54,9 +75,7 @@ const PosterViewItem = ({ isSeries = false, item }: Props) => {
         >
           {showUnwatchedCount && (
             <div className="absolute top-0 right-0 flex min-w-[2.81rem] justify-center rounded-bl-md bg-panel-background-overlay p-3 font-semibold opacity-100 transition-opacity group-hover:opacity-0">
-              {unwatchedCount || (
-                <Icon path={mdiCheckboxMarkedCircleOutline} size={1} className="text-panel-icon-important" />
-              )}
+              {topRightIcon}
```
hidden4003:
```suggestion
 {unwatchedCount || (episodeCount ? (
  <Icon path={mdiCheckboxMarkedCircleOutline} size={1} className="text-panel-icon-important" />
) : (
  <Icon path={mdiFileDocumentAlertOutline} size={1} className="text-panel-icon-warning" />
))}
```

src/components/SeriesPoster.tsx
```diff
@@ -36,6 +42,26 @@ const SeriesPoster = (props: Props) => {
...
+  } = useRefreshAniDBSeriesMutation();
+
+  const createSeries = (anidbId: number, event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
+    event.preventDefault();
+
+    refreshSeries({ anidbID: anidbId, immediate: true, createSeriesEntry: true, force: true }, {
+      onSuccess: () => {
+        toast.success('Series added successfully!');
+        invalidateQueries(['series', 'without-files']);
```
hidden4003: @harshithmohan  don't we also need to invalidate `['series', seriesId, 'related']` and `['series', seriesId, 'similar']`?

hidden4003 (changes requested):

src/components/Collection/PosterViewItem.tsx
hidden4003: Also forgot to mention we need tool tips on the icons too as it is not immediately obvious what the yellow one means.

src/components/SeriesPoster.tsx
harshithmohan: We would need to invalidate the related and similar anime for the series we are on, we don't have access to *that* series ID in here. We only have access to the series we are creating.

1. When this is called from dashboard, signalr takes care of invalidating
2. When called from series page, if we want to invalidate only related and similar, we would need to pass the seriesId to this component but I feel that's not worth it. We could just invalidate everything related to series `invalidateQueries(['series'])` or just not do anything

src/components/SeriesPoster.tsx
```diff
@@ -53,6 +79,18 @@ const SeriesPoster = (props: Props) => {
           </div>
         )}
 
+        {(isAnidb && anidbSeriesId) && (
+          <div className="pointer-events-none absolute z-15 flex h-full p-3 opacity-0 transition-opacity group-hover:pointer-events-auto group-hover:opacity-100">
+            <Button
+              className="pointer-events-auto h-fit"
+              onClick={event => createSeries(anidbSeriesId, event)}
+              tooltip="Add to collection"
+            >
```
harshithmohan, line 97: Add loading state to this based on the pending status of the mutation

src/components/SeriesPoster.tsx
```diff
@@ -92,10 +130,10 @@ const SeriesPoster = (props: Props) => {
     );
   }
 
-  if (anidbSeriesId ?? anidbEpisodeId) {
+  if (anidbEpisodeId ?? anidbSeriesId) {
     return (
       <a
-        href={`https://anidb.net/${anidbSeriesId ? `anime/${anidbSeriesId}` : `episode/${anidbEpisodeId}`}`}
+        href={`https://anidb.net/${anidbEpisodeId ? `episode/${anidbEpisodeId}` : `anime/${anidbSeriesId}`}`}
```
harshithmohan, line 145: Why was this changed?

src/pages/dashboard/components/EpisodeDetails.tsx
```diff
@@ -89,6 +89,7 @@ const EpisodeDetails = ({ episode, isInCollection = false, showDate = false }: P
         title={episode.SeriesTitle}
         subtitle={title}
         shokoId={episode.IDs.ShokoSeries}
+        anidbSeriesId={episode.IDs.Series}
```
harshithmohan, line 92: `anidbSeriesId` is only needed for creating a series, doesn't make sense to pass it when rendering the poster for an episode

src/components/Collection/PosterViewItem.tsx
harshithmohan: `mdiFileDocumentRemoveOutline` might be better than `mdiFileDocumentAlertOutline`. Keyword: "might"

Image

src/components/SeriesPoster.tsx
harshithmohan: Oh, I understand why it was changed but doesn't need it if seriesId is not passed for episodes

src/components/SeriesPoster.tsx
hidden4003: Right, in case 2 we need to invalidate the series that is being viewed not the one that component is trying to display data for. Either prop or context but sounds too much work. How bad is the blast radius to just invalidate everything `series`? Since it should be rarely used should be ok.

src/components/Collection/PosterViewItem.tsx
pokyunn: I agree, but should I change In Utilities nav too? As its the same type "series without files" to keep them easy to associate?

src/pages/dashboard/components/EpisodeDetails.tsx
pokyunn: Its needed because "upcomming anime" shows per-epsode posters, to be able to grab series there its needed to pass the anidbSeriesId

src/components/SeriesPoster.tsx
pokyunn: using `['series']` it does fix, but the image "fails load" (not refreshed) but clicking in it to go to new page, it show the image

src/components/SeriesPoster.tsx
pokyunn: Instead I choose to hide this button and change the text
`{isRefreshPending ? 'Fetching from AniDB' : 'View on AniDB'}`

src/pages/dashboard/components/EpisodeDetails.tsx
harshithmohan: So it does 😔

src/components/SeriesPoster.tsx
hidden4003: Our button already has a `loading` prop you just need to use that and it will take care of showing that it is loading.

src/components/SeriesPoster.tsx
hidden4003: Similar/related section does not update since you are invoking code inside the poster and series page or dashboard does not know you did that. For the information to update properly you need a context or a callback prop to pass back the event that series was added for queries to be re-triggered in the whatever page poster is in.

src/components/Collection/PosterViewItem.tsx
```diff
@@ -53,10 +53,18 @@ const PosterViewItem = ({ isSeries = false, item }: Props) => {
...
-                <Icon path={mdiCheckboxMarkedCircleOutline} size={1} className="text-panel-icon-important" />
-              )}
+            <div
+              data-tooltip-id="tooltip"
+              data-tooltip-content={(() => {
+                if (unwatchedCount) return 'Unwatched files';
+                if (episodeCount) return 'All files have been watched';
+                return 'Series Without Files';
+              })()}
+              className="absolute top-0 right-0 z-10 flex min-w-[2.81rem] justify-center rounded-bl-md bg-panel-background-overlay p-3 font-semibold opacity-100"
```
hidden4003: you can put tooltip on the icon, then you don't need to split the logic

pokyunn: Yes, but in this case, still need for tooltip in this div for "unwatched epsode number", and linter will expand the Icons in multiple lines due to size, 

If its not a problem, I change

src/components/SeriesPoster.tsx
harshithmohan: > using `['series']` it does fix, but the image "fails load" (not refreshed) but clicking in it to go to new page, it show the image

Image does not load because the image does not exist yet. By the time the data is refreshed from the API, the server hasn't downloaded the image yet. That is expected. You cannot fix it. The image might be downloaded the next second or the next hour.

src/components/Collection/PosterViewItem.tsx
harshithmohan: We don't need a tooltip for the number. That's a standard way to show it

pokyunn: ok

src/components/SeriesPoster.tsx
pokyunn: This loading put this icon image over te content of button, since Im using image they blend and feel like "nothing happen"

So for me, hidding butto,
And put text OR loading like in collection feels "more real"
Screenshot_20260718_121626

Keep your sugestion to use button prop or can I use this loading?

harshithmohan: Then I would suggest the following
1. Just put the loading icon in the middle of the poster and with colour `text-panel-text-primary`. Remove the AniDB logo as well.
2. Set the disabled prop on the button when it's pending so that you can't click it again.

@hidden4003 that good?

hidden4003: > Then I would suggest the following
> 
>     1. Just put the loading icon in the middle of the poster and with colour `text-panel-text-primary`. Remove the AniDB logo as well.
> 
>     2. Set the disabled prop on the button when it's pending so that you can't click it again.
> 
> 
> @hidden4003 that good?

yea that is fine with me

pokyunn: ok with it

src/components/Collection/PosterViewItem.tsx
harshithmohan: > I agree, but should I change In Utilities nav too? As its the same type "series without files" to keep them easy to associate?

Already using it for "Missing Episodes" in the other PR and this is technically missing episodes so I think it's fine. Also why we have a tooltip. If you think it doesn't make sense you can make it back to `mdiFileDocumentAlertOutline`. I don't mind it either way. I didn't realize we were already using that icon when I commented about it first

src/components/Collection/PosterViewItem.tsx
```diff
@@ -53,18 +53,26 @@ const PosterViewItem = ({ isSeries = false, item }: Props) => {
...
+                    data-tooltip-content="All files have been watched"
+                    path={mdiCheckboxMarkedCircleOutline}
+                    size={1}
+                    className="text-panel-icon-important"
+                  />
+                )
+                : (
+                  <Icon
+                    data-tooltip-id="tooltip"
+                    data-tooltip-content="Series Without Files"
```
harshithmohan: Put it in a sentence like "Series does not have files" or "Series does not have episodes"

src/components/Layout/TopNav.tsx
```diff
@@ -306,7 +306,7 @@ const TopNav = () => {
               text="Release Management"
             />
             <LinkMenuItem
-              icon={mdiFileDocumentAlertOutline}
+              icon={mdiFileDocumentRemoveOutline}
```
harshithmohan: Image

Now revert it 😔

harshithmohan: The poster can use either but topnav cannot

src/components/SeriesPoster.tsx
```diff
@@ -48,8 +75,27 @@ const SeriesPoster = (props: Props) => {
...
+                </>
+              )}
+          </div>
+        )}
+
+        {isAnidb && anidbSeriesId && (
+          <div className="pointer-events-none absolute z-15 flex h-full p-3 opacity-0 transition-opacity group-hover:pointer-events-auto group-hover:opacity-100">
+            <Button
+              disabled={isRefreshPending}
+              className="pointer-events-auto h-fit"
```
harshithmohan: Why does it need these classes?

harshithmohan (changes requested):

src/components/Collection/PosterViewItem.tsx
```diff
@@ -53,10 +53,26 @@ const PosterViewItem = ({ isSeries = false, item }: Props) => {
           zoomOnHover
         >
           {showUnwatchedCount && (
-            <div className="absolute top-0 right-0 flex min-w-[2.81rem] justify-center rounded-bl-md bg-panel-background-overlay p-3 font-semibold opacity-100 transition-opacity group-hover:opacity-0">
-              {unwatchedCount || (
-                <Icon path={mdiCheckboxMarkedCircleOutline} size={1} className="text-panel-icon-important" />
-              )}
+            <div className="absolute top-0 right-0 z-10 flex min-w-[2.81rem] justify-center rounded-bl-md bg-panel-background-overlay p-3 font-semibold opacity-100">
```
hidden4003: The watched-status badge previously faded out on hover (`opacity-100 transition-opacity group-hover:opacity-0`), matching the `groupCount` badge lower in the same file, which still has that behavior.

This PR dropped `transition-opacity group-hover:opacity-0` from the watched-status badge, so it now stays permanently visible on hover, while the group-count badge still fades away — two visually similar corner badges in the same component now behave differently on hover.

**Ask:** was this intentional (e.g., to keep the tooltip readable while hovering)? If so, worth a short comment explaining why; if not, might want to restore the fade for consistency with the sibling badge.

hidden4003:
# PR #1395 — "Improve series poster" — Status Summary

_Author: pokyunn · Reviewers: hidden4003, harshithmohan · Status: OPEN, CHANGES_REQUESTED (as of 2026-07-19)_

## ✅ Closed / Resolved

- **Add-to-collection shortcut on posters** — `createSeries` handler + button added in `SeriesPoster.tsx`, using `useRefreshAniDBSeriesMutation`.
- **Loading/disabled state while adding a series** — spinner centered over poster (AniDB logo/text removed while pending), button `disabled` while mutation is pending. Confirmed by both reviewers.
- **`anidbSeriesId` prop threading** through `EpisodeDetails.tsx` / `SeriesDetails.tsx` — needed so per-episode posters (dashboard "upcoming anime") can resolve the series to add.
- **Watched / unwatched / no-files indicator icon + tooltips** in `PosterViewItem.tsx`.
- **Tooltip scope** — only on the status icons, not on the raw unwatched-episode count (confirmed as unnecessary by harshithmohan).
- **Query invalidation approach** — settled on a broad `invalidateQueries(['series'])` instead of targeting `related`/`similar` keys individually; accepted as acceptable blast radius since the action is rare.
- **Image-not-refreshing-immediately** — clarified as expected (server hasn't downloaded the new image yet at mutation completion), not fixable, not a real bug.
- **Add-to-collection button z-index vs. `overlayOnHover` darkening layer** — flagged as a possible click-through risk (`z-15` button vs. `z-50` shared overlay), but author confirmed the button works as expected in practice. No action needed.

## ⏳ Open

1. **`TopNav.tsx:309`** — revert nav icon back to `mdiFileDocumentAlertOutline`. Poster can keep `mdiFileDocumentRemoveOutline`, but TopNav should not use it.
2. **`PosterViewItem.tsx:70`** — reword tooltip "Series Without Files" into a full sentence, e.g. "Series does not have files."
3. **Inconsistent hover-fade on poster badges in `PosterViewItem.tsx`** — watched-status badge no longer fades on hover (`transition-opacity group-hover:opacity-0` was dropped), while the sibling `groupCount` badge in the same file still fades. Needs a decision on whether this was intentional.

## Next step
Address items 1–2 (small/mechanical), get confirmation from the author on item 3 (intentional or not), then re-request review.

src/components/Layout/TopNav.tsx
pokyunn: reverted

src/components/SeriesPoster.tsx
pokyunn: I used src/components/Collection/PosterViewItem.tsx line 78, aka  'Edit Series' / 'Edit Group' button, as reference, but I'm not good with frontend, so just changed until it show the button, so I'm ok with any changes to the classes as I dont really know how to choose them 

But starting from the reference and adding `absolute z-15` and removing `bg-panel-background-poster-overlay` made able to click it, as before it was triggering background link

src/components/Collection/PosterViewItem.tsx
pokyunn: I'm not good with frontend, but when hovering mouse over the icons, it triggered "zoom in and the layer with edit series button" making tooltip visible but not the icon from it 
image
So it  was intentional
vs keeping it visible
image

I agree its inconsistent, but its out off my frontend skills, so I accept any suggestion

hidden4003: Tooltip should not be visibile if icon is not visible.

pokyunn: In this case tooltip will never be displayed, because on hovering icons will trigger background, defeating tooltip purpose.
How can I prevent icons to trigger this background effect?

hidden4003: I tried on stable if I remove `transition-opacity group-hover:opacity-0` then icon does not disappear, try removing `z-10` it should not be needed here.

harshithmohan:
https://github.com/user-attachments/assets/bf48caad-a5c9-4265-86dc-569d3fde1ab4

@pokyunn Pushed few changes because it was faster than commenting, also changes behaviour of spinner to show up without hover.
With the latest server daily, there is now a way better chance of the poster loading, apparently we were caching the 404 response before

pokyunn:
> screen_recording_20260720-2017.mp4 
> @pokyunn Pushed few changes because it was faster than commenting, also changes behaviour of spinner to show up without hover. With the latest server daily, there is now a way better chance of the poster loading, apparently we were caching the 404 response before

No problem, thanks, just noticed after this change it has a small "blink" in collection PosterViewItem when hovering icon

harshithmohan:
I accidentally removed the `z-10`

PR #1397, opened 2026-07-20
fix: reset modal inputs on close

DrSilverFox:
Resets the modal inputs when closed as otherwise the respective fields are filled with previous values when the modal is reopened again. This issue was found in two places:
* Series without files - Add series modal
* File renamer - Add files modal

### DIFF
--- a/src/components/Utilities/Renamer/AddFilesModal.tsx
+++ b/src/components/Utilities/Renamer/AddFilesModal.tsx
@@ -98,10 +98,16 @@ const AddFilesModal = ({ onClose, show }: Props) => {
   );
   const [series, seriesCount] = useFlattenListResult(seriesQuery.data);
 
+  const handleCancel = () => {
+    setSearch('');
+    setPageSize(10);
+    onClose();
+  };
+
   return (
     <ModalPanel
       show={show}
-      onRequestClose={onClose}
+      onRequestClose={handleCancel}
       header="Add Files"
       size="sm"
       noGap
--- a/src/components/Utilities/Renamer/AddFilesModal.tsx
+++ b/src/components/Utilities/Renamer/AddFilesModal.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { mdiMagnify, mdiPlayCircleOutline } from '@mdi/js';
 import { Icon } from '@mdi/react';
 import { toNumber } from 'lodash';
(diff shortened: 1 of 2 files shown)


src/components/Utilities/SeriesWithoutFiles/AddSeriesModal.tsx
```diff
@@ -24,6 +24,11 @@ const AddSeriesModal = ({ onClose, show }: Props) => {
 
   const searchQuery = useSeriesAniDBSearchQuery(debouncedSearch, !!debouncedSearch);
 
+  const handleCancel = () => {
+    setSearchText('');
+    onClose();
+  };
+
```
hidden4003:
```suggestion
  // Reset fields on open
  useEffect(() => {
    setSearchText('');
  }, []);
```

src/components/Utilities/Renamer/AddFilesModal.tsx
```diff
@@ -98,10 +98,16 @@ const AddFilesModal = ({ onClose, show }: Props) => {
   );
   const [series, seriesCount] = useFlattenListResult(seriesQuery.data);
 
+  const handleCancel = () => {
+    setSearch('');
+    setPageSize(10);
+    onClose();
+  };
+
```
hidden4003, line 105: No need for this just use `useEffect`

```suggestion
  // Reset fields on open
  useEffect(() => {
    setSearch('');
    setPageSize(10);
  }, []);
```

hidden4003 (changes requested):

src/components/Utilities/Renamer/AddFilesModal.tsx
harshithmohan: deps array should contain `show` so that it resets everytime it's hidden/shown
```
  useEffect(() => {
    setSearch('');
    setPageSize(10);
  }, [show]);
```

src/components/Utilities/SeriesWithoutFiles/AddSeriesModal.tsx
harshithmohan: deps array should contain `show` so that it resets everytime it's hidden/shown
```
  useEffect(() => {
    setSearchText('');
  }, [show]);
```

src/components/Utilities/Renamer/AddFilesModal.tsx
DrSilverFox: I had gone with the common function as when I had looked at a couple of other files to get myself familiar with the code they were also doing the same thing.

PR #1402, opened 2026-07-22
fix: add force option to series TMDB refresh action

Foowy:
## Summary
- `useRefreshSeriesTMDBInfoMutation` posted an empty body to `Series/{id}/TMDB/Show|Movie/Action/Refresh`, so `Force` always defaulted to `false` server-side — the "Update TMDB Info" button in the series edit panel never actually forced a refresh (only reachable via Swagger).
- Added a `force` param threaded through to both `Action/Refresh` calls, and a second "Update TMDB Info - Force" button, mirroring the existing AniDB force-refresh pattern.

**Back to draft — reviewing scope.** Server-side investigation turned up that the need for a separate Force button may itself be a symptom: `UpdateShow`/`UpdateMovie` gate a full refresh behind a 14-day TMDB Changes-API cache that's proven unreliable for translation/title-only edits (the root cause of shows getting stuck with placeholder episode titles that never resolve on a normal refresh). Considering just removing that 14-day incremental-cache limit server-side instead of exposing a client-side Force escape hatch for it — TMDB has no strict rate limit that requires the optimization, and it's been causing more correctness issues than the API calls it was saving. If that lands, this PR's Force button may become unnecessary (a plain refresh would always be complete) and should be reconsidered before merging.

## Test plan
- [x] `dprint fmt` / `oxlint` clean on changed files
- [x] Manual verification against a running Shoko Server instance

cc @harshithmohan — would appreciate a review here.

### DIFF
--- a/src/components/Collection/Series/EditSeriesTabs/UpdateActionsTab.tsx
+++ b/src/components/Collection/Series/EditSeriesTabs/UpdateActionsTab.tsx
@@ -26,6 +26,10 @@ const UpdateActionsTab = ({ seriesId }: Props) => {
     updateTmdbImagesMutation({ force: true });
   };
 
+  const triggerTmdbRefresh = (force: boolean) => {
+    refreshTmdb({ force });
+  };
+
   return (
     <div className="flex h-88 grow flex-col gap-y-4 overflow-y-auto">
       <Action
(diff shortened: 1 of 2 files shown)


Foowy:
My qualm is adding yet another button/link. Should force only be done via swagger only, the refresh button should always be forceful, or is adding another clickable link best option?

Foowy:
Closing this in favor of a server-side fix: ShokoAnime/ShokoServer#1393 shrinks the TMDB incremental changes cache window from 14 days to 1, which addresses the same "stuck placeholder titles" problem this Force button was working around, without needing a client-side escape hatch.

PR #1414, opened 2026-07-26
Add Orval + Zod runtime API response validation

hidden4003:
## Summary
- Adds [Orval](https://orval.dev) to generate TypeScript types + Zod schemas from Shoko Server's live OpenAPI v3 spec into `src/core/api/generated/` (committed; regenerate locally via `pnpm orval` against a running server, since CI has none).
- Adds `validateResponse()` / `SchemaValidationError` (`src/core/api/validateResponse.ts`) so a response that doesn't match its schema throws instead of letting malformed data silently reach components as `undefined` fields.
- `queryClient.ts` surfaces a `SchemaValidationError` as an immediate "Invalid API Response" toast, skipping the existing 4x retry (retrying won't fix a shape mismatch).
- Pilot conversion of `tag` and the tag-related `series` hooks to prove the pattern; the rest of `src/core/react-query/` is intentionally left for incremental follow-up.
- Along the way this caught two real, pre-existing bugs: `TagType.Source` was typed as a closed `'AniDB' | 'User'` union when the server doesn't actually constrain it, and a `cleanString()` helper only guarded against `undefined`, not `null` (which the real schema allows for `Description`).

Draft because this establishes a new pattern (Orval + Zod validation, a `src/core/api/generated` directory excluded from oxlint, a few workarounds for upstream Orval/spec quirks) that I'd like other maintainers to weigh in on before it's adopted more broadly.

## Test plan
- [x] `pnpm dprint`, `pnpm oxlint`, `pnpm stylelint`, `pnpm tscheck` all pass
- [x] `pnpm orval` runs cleanly against a live Shoko Server dev instance
- [x] `pnpm start` boots and proxies correctly
- [ ] Manual click-through of Collection list view and a series detail page's tags panel (please verify — I didn't have a browser automation tool available in my sandbox)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


harshithmohan:
So right now we only generate types and not the hooks also?

hidden4003:
> So right now we only generate types and not the hooks also?

Our hooks have custom logic in them which is why I did not set up hook generation as well, I can add it if you want to see how it looks.

PR #1425, opened 2026-08-10
Rework collection filter format to align with server, add preset editing

hidden4003:
## Summary
- Filter state moves from a flat map to a recursive AND/OR/NOT tree matching the shape of the server's expression format, enabling arbitrary nesting and negation (not just flat AND) instead of the old implicit-AND-only model.
- Saved filter presets can now be loaded back into the sidebar and edited in place ("Edit Filter" action), then saved back to the existing preset — previously editing was disabled entirely since there was no way to parse a saved condition tree back into UI state.
- Conditions the UI can't represent (date/comparison-based expressions) are preserved read-only rather than dropped, so editing part of a preset never corrupts the rest.
- Sidebar UI stays visually unchanged for the common flat-AND case; AND/OR/NOT controls only appear when actually needed.

### DIFF
--- a/docs/FILTERS.md
+++ b/docs/FILTERS.md
@@ -1,0 +1,330 @@
+# Collection Filter System
+
+Developer-level specification for the Collection filter builder
+(`src/components/Collection/Filter/`) and its supporting layers
+(`src/core/utilities/filterTree.ts`, `src/core/slices/collection.ts`,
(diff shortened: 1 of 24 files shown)


harshithmohan (commented):
There was a review comment here, which was wrong

PR #1431, opened 2026-08-15
Edit Release Info modal

harshithmohan:
## Summary

Adds an "Edit Release Info" modal to the link-with-providers workflow, letting users edit the release metadata (series/episode mapping, version, source, chaptered, creditless, comment) of one or more selected files before submitting them.

### Edit Release Info modal
- **Entry point** — the previously-disabled "Edit Release Info" menu button is now enabled (`E` hotkey), gated by the same condition as "Search for Release Info": all selected links must be in `ready`/`init`.
- **Two-step flow** — a series-search screen (shared `AnimeSelectPanel`, debounced AniDB search) when no series is selected, then a review form once a series is chosen (or when bulk-selecting mixed series).
- **Series selection** — supports refresh and force-refresh of AniDB data; selecting an "unknown" series type fetches it first.
- **Episode selection** — lists the series' episodes plus an "Auto-match (from filename)" option (`AUTO_MATCH_EPISODE_ID`); per-file auto-match resolves the real episode at save time.
- **Fields** — version, source, chaptered, creditless, and comment; clearing the version input falls back to empty rather than `NaN`.
- **Save gating** — Save stays disabled until a field is touched ~~and, for a single-series edit, until an episode (or auto-match) is selected~~.
- **Bulk / mixed values** — when selected files disagree, the UI reflects it: source shows a disabled `-`, version renders empty, chaptered/creditless show an indeterminate checkbox, and the episode selector shows a disabled "Multiple episodes selected" entry.

### Shared component changes
- **Checkbox** — new `indeterminate` prop (sets the native `input.indeterminate` DOM property, which React doesn't expose, and renders a half-filled icon).
- **SelectEpisodeList** — new `standalone` variant and an optional `number` (the auto-match option has no episode number).
- **AnimeSelectPanel** — extracted from `LinkFilesTab` into a shared component reused by the modal; API changed from `changeSelectedSeries`/`seriesUpdating` to `onSelect`/`showLoading`.

### Save path (`LinkFilesWithProviders.handleSaveReleaseInfo`)
For each selected link: merges the release patch, resolves a cross-reference (auto-match resolves via `detectShow` → matching episode), writes `CrossReferences` with a 0–100% range, appends the `+User` marker to `ProviderName`, and marks the link `ready`. Links whose auto-match fails keep their previous state (no `ready` transition) and are counted in an error toast.

### Modal state (`useReleaseInfoForm` hook)
Form state, init-from-selection logic, and touched-field tracking are encapsulated in a new hook under `src/hooks/utilities`:
- Form fields derive from `ReleaseInfoType` via indexed access types.
- Touched-field names are compile-checked via a `TouchableField` union (in `link-files-with-providers.ts`).
- The save patch is built declaratively through a typed `setIfTouched` helper.
- Mixed-selection flags are consolidated in a single `hasDifferent` Immer object state.
- Init runs from a `useEffectEvent` gated on `show` only, so the form doesn't reset when the page re-renders while the modal is open.

### Types
- Added `CrossReferenceType` and `TouchableField` to `link-files-with-providers.ts`.
- Added the `AUTO_MATCH_EPISODE_ID = -1` sentinel to ~~`auto-match-logic.ts`~~ `releaseInfoHelpers.ts`.

## Changes after review

- **Version/Source empty handling** — the save patch now uses an explicit `formState.version === '' ? undefined : formState.version` (and the same for `source`) instead of `||`. This preserves a typed `0` (accepted by the server) while still dropping the empty-string sentinel.
- **Shared constants** — `EDITABLE_STATES` and `AUTO_MATCH_EPISODE_ID` now live in `releaseInfoHelpers.ts`; the `Menu` component reuses `EDITABLE_STATES` instead of an inline `['ready', 'init']` literal.
- **Auto-matched specials** — special files with no number in the filename are spread across consecutive special episodes (S1, S2, S3…) via a per-file `specials` counter, matching `LinkFilesTab`'s `autoFill` behavior.
- **Series-name suggestion** — `AnimeSelectPanel` now syncs its search text when the `placeholder` prop changes, so the auto-detected series-name suggestion reaches the search box.
- **Differing episodes** — bulk-editing files with distinct episode links no longer forces an episode selection; the selector shows "Multiple episodes selected" and existing per-file links are preserved unless an episode (or auto-match) is explicitly chosen.

## Verification
- `pnpm lint` passes (dprint + oxlint + stylelint).
- `tsc --noEmit` clean.


src/pages/utilities/LinkFilesWithProviders.tsx
```diff
@@ -178,16 +188,79 @@ const LinkFilesWithProviders = () => {
...
+        Object.assign(draft[link.id].release, releaseInfo);
+
+        let matchFailed = false;
+        if (crossReference) {
+          const { episodeId: selectedEpisodeId, episodes, seriesId } = crossReference;
+          let episodeId = selectedEpisodeId;
+          if (episodeId === AUTO_MATCH_EPISODE_ID) {
+            const details = detectShow(link.file?.Locations?.[0]?.RelativePath);
+            if (details) {
+              const episodeNumber = details.episodeType === EpisodeTypeEnum.Special && details.episodeStart === 0
```
hidden4003, line 220: Auto-match for bulk-selected special episodes always resolves to episode number 1 instead of incrementing per file.

User selects multiple Special-type files with no number in the filename, opens Edit Release Info, picks the series, and leaves episode as "Auto-match (from filename)". `handleSaveReleaseInfo` computes `episodeNumber = 1` for every file (no per-file counter, unlike the removed `specials += 1` logic in LinkFilesTab.tsx's `autoFill`), so distinct special files are all cross-referenced to the same 'Special Episode 1' instead of being spread across S1, S2, S3…

src/hooks/utilities/useReleaseInfoForm.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    setTouchedFields(new Set());
+
+    const allSame = (selector: (release: ReleaseInfoType) => unknown) =>
+      selectedLinks.every(link => selector(link.release) === selector(first));
+
+    const hasMultipleSeries = isBulk && !allSame(link => link.CrossReferences[0]?.AnidbAnimeID);
+    const initialSeriesId = hasMultipleSeries ? undefined : first.CrossReferences[0]?.AnidbAnimeID;
+
+    setInitialSeriesName(
```
hidden4003, line 50: Auto-detected series-name suggestion is set via useEffect one render too late, so AnimeSelectPanel never picks it up as its initial placeholder.

Modal opens for files with no existing series link: on the render where `show` flips to true, `AnimeSelectPanel` mounts with `placeholder={initialSeriesName}`, which is still stale (empty or leftover from a prior session, since the modal never unmounts). Only afterward does the `initForm` effect call `setInitialSeriesName(findMostCommonShowName(...))`, but `AnimeSelectPanel`'s internal `useState(placeholder)` only consumes its argument on first render, so the corrected guess never reaches the search box.

src/components/Utilities/Unrecognized/LinkFilesWithProvider/EditReleaseInfoModal.tsx
```diff
@@ -0,0 +1,389 @@
...
+    ) => {
+      if (!touchedFields.has(field)) return;
+      if (value === undefined) return;
+      releaseInfo[field] = value;
+    };
+
+    setIfTouched('Version', formState.version || undefined);
+    setIfTouched('IsChaptered', formState.isChaptered);
+    setIfTouched('IsCreditless', formState.isCreditless);
+    setIfTouched('Source', formState.source || undefined);
```
hidden4003: Version's `formState.version || undefined` and Source's `formState.source || undefined` use identical `||` syntax for two different reasons, only one of which is functionally required — the pattern reads as consistent but isn't, which invites a future "fix" that breaks Source.

Source's `''` (the disabled placeholder option's value) is not a real `ReleaseSource` enum member, so `||` is load-bearing there: it coerces the sentinel `''` to `undefined` before `setIfTouched` runs, and switching it to `??` would let an invalid empty string leak into `releaseInfo.Source`. Version's `||` has no equivalent justification — `formState.version` is a number, the input has `min={1}` (unenforced without a native form submit, so 0 is reachable by typing), and there is no domain concept of version 0, so coercing a falsy 0 to undefined has no real consequence either way. Because both lines look the same, a future refactor applying CLAUDE.md's "prefer `??`" rule uniformly across both would silently reintroduce the Source bug while \"fixing\" a Version line that was never actually broken. Recommend making the two cases visually distinct, e.g. an explicit `formState.source === '' ? undefined : formState.source` for Source (documenting that `''` is the sentinel) versus leaving Version as-is or switching it to `??` (functionally equivalent for Version, and correctly signals that no sentinel-stripping is happening there).

src/components/Utilities/Unrecognized/LinkFilesWithProvider/EditReleaseInfoModal.tsx
```diff
@@ -0,0 +1,389 @@
...
+    const setIfTouched = <FieldName extends TouchableField>(
+      field: FieldName,
+      value?: ReleaseInfoType[FieldName],
+    ) => {
+      if (!touchedFields.has(field)) return;
+      if (value === undefined) return;
+      releaseInfo[field] = value;
+    };
+
+    setIfTouched('Version', formState.version || undefined);
```
hidden4003: `formState.version || undefined` violates the project's `??` convention for non-boolean values.

src/hooks/utilities/useReleaseInfoForm.ts
```diff
@@ -0,0 +1,103 @@
...
+      initialSeriesId
+        ? ''
+        : findMostCommonShowName(selectedLinks.map(link => detectShow(link.file?.Locations?.[0]?.RelativePath))),
+    );
+
+    const hasDifferentEpisodes = isBulk && !allSame(link => link.CrossReferences[0]?.AnidbEpisodeID);
+    const initialEpisodeId = initialSeriesId && !hasDifferentEpisodes
```
hidden4003, line 57: Bulk-editing files with differing episodes forces the user to pick an episode to enable Save, which then overwrites every file's correct existing episode link as a side effect of an unrelated edit.

User selects several same-series files with distinct, already-correct episode links (`hasDifferentEpisodes` true, `initialEpisodeId` undefined) and only wants to bump `Version`. Save stays disabled (`disabled={... || (!!formState.selectedSeriesId && !formState.selectedEpisodeId)}`) until an episode is chosen. Picking one episode homogenizes all selected files to it; picking Auto-match re-derives each from filename, potentially diverging from a previously hand-corrected link — either way, correct per-file episode assignments are silently overwritten.

src/components/Utilities/Unrecognized/LinkFilesWithProvider/Menu.tsx
```diff
@@ -11,34 +11,39 @@ type Props = {
...
     linkCount,
+    onEditReleaseInfo,
     openSearchDialog,
     removeLinks,
     selectedLinks,
     toggleAllSelectedLinks,
   } = props;
 
+  const releaseActionsDisabled = !selectedLinks.length
+    || !selectedLinks.every(link => ['ready', 'init'].includes(link.state));
```
hidden4003: Menu re-implements the editable-state check with an inline array literal instead of reusing the shared EDITABLE_STATES set, risking drift.

A future change updates `EDITABLE_STATES` in LinkFilesWithProviders.tsx but the developer misses this duplicated `['ready', 'init']` literal in Menu.tsx, causing menu buttons' enabled/disabled state to disagree with what the actual handlers allow.

hidden4003 (changes requested):

hidden4003 (approved):

PR #1436, opened 2026-08-17
feat(avdump): Display CRC32 filename matching status

pokyunn:
Introduces a visual warning for files where the computed CRC32 hash does not match or is absent from the filename. The default 'ready to dump' state now explicitly uses a distinct icon and tooltip to confirm a successful CRC32 match.

### DIFF
--- a/src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
+++ b/src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
@@ -3,8 +3,9 @@ import type { MouseEvent } from 'react';
 import {
   mdiDumpTruck,
   mdiFileDocumentAlertOutline,
+  mdiFileDocumentArrowRightOutline,
   mdiFileDocumentCheckOutline,
-  mdiFileDocumentPlusOutline,
+  mdiFileDocumentOutline,
   mdiFileDocumentRefreshOutline,
   mdiLoading,
 } from '@mdi/js';
@@ -50,7 +51,7 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
 
     if (dumpSession?.status === 'Success') {
       return {
-        path: mdiFileDocumentCheckOutline,
+        path: mdiFileDocumentArrowRightOutline,
         color: 'text-panel-icon-important',
         title: 'Dumped Successfully!',
         state: 'success',
@@ -68,7 +69,7 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
 
     if (file.AVDump.LastDumpedAt) {
       return {
-        path: mdiFileDocumentCheckOutline,
+        path: mdiFileDocumentArrowRightOutline,
         color: 'text-panel-icon-important',
         title: 'Previously Dumped!',
         state: 'success',
@@ -84,19 +85,26 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
       } as const;
     }
 
-    if (!file.Locations[0].RelativePath?.includes(file.Hashes.find(hash => hash.Type === 'CRC32')?.Value ?? '')) {
-      return {
-        path: mdiFileDocumentAlertOutline,
-        color: 'text-panel-text-warning',
-        title: 'The computed CRC32 does not match or is not present in filename!',
-        state: 'info',
-      } as const;
+    const crc32 = (/\b[0-9A-F]{8}\b/i.exec(file.Locations[0].RelativePath))?.[0].toUpperCase();
+    if (crc32) {
+      return crc32 === file.Hashes.find(hash => hash.Type === 'CRC32')?.Value
+        ? {
+          path: mdiFileDocumentCheckOutline,
+          color: 'text-panel-text-primary',
+          title: 'Computed CRC32 matches filename!',
+          state: 'info',
+        } as const
+        : {
+          path: mdiFileDocumentAlertOutline,
+          color: 'text-panel-text-warning',
+          title: 'Computed CRC32 does not match filename!',
+          state: 'info',
+        } as const;
     }
 
     return {
-      path: mdiFileDocumentPlusOutline,
-      color: 'text-panel-text-primary',
-      title: 'File CRC32 is present in filename and matches',
+      path: mdiFileDocumentOutline,
+      color: 'text-panel-text',
       state: 'info',
     } as const;
   }, [file, dumpSession, truck]);
--- a/src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
+++ b/src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
@@ -17,6 +17,7 @@ import { useAvdumpFilesMutation } from '@/core/react-query/avdump/mutations';
 import { useSelector } from '@/core/store';
 import toast from '@/core/toast';
 import { copyToClipboard, processError } from '@/core/util';
+import { Crc32Regex } from '@/core/utilities/auto-match-regexes';
 import getEd2kLink from '@/core/utilities/getEd2kLink';
 
 import type { FileType } from '@/core/types/api/file';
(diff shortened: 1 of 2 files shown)


harshithmohan:
@hidden4003 @revam  What do you think?
I think we should only check for CRC if there is one present, maybe using regex to get it from the file, or maybe that is something the server can check if it matches. Either works but I don't think it's correct to show "CRC not found in the filename" as a warning because it is common to have files without CRC in the filename

pokyunn:
Here the idea is to help preventing mismatch/corrupt files to be avdumped and added to anidb (as this icon is only in unrecognized files utility).
Is possible with regex something like \b[0-9a-fA-F]{8}\b to only match against pattern
But I think mostly files without crc come from *Arss (and if not match anidb they go manual import and should not be avdumped) or custom renamers (after import and, this icon is not shown in files search) 

> @hidden4003 @revam What do you think? I think we should only check for CRC if there is one present, maybe using regex to get it from the file, or maybe that is something the server can check if it matches. Either works but I don't think it's correct to show "CRC not found in the filename" as a warning because it is common to have files without CRC in the filename

da3dsoul:
I think it should only show an indicator when they are found and don't match. You can assume good or unknown otherwise

revam:
> I think it should only show an indicator when they are found and don't match. You can assume good or unknown otherwise

@da3dsoul stole my lines before I got here, and also delivered them better. what he said.

harshithmohan:
So... regex? *insert thisisfine gif*

revam:
> So... regex? _insert thisisfine gif_

Regular expressions is the solution to all worldly problems. Yes. You can even play doom through them now, last I checked.

hidden4003:
First of all `.includes()` is case sensitive and I don't know if relative path is transformed to lower/upper case.
I don't see how you are going to avoid regex if you need to detect a crc like pattern, detecting patters is what regexes are made for.

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
```diff
@@ -27,7 +28,7 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
   const fileId = file.ID;
   const dumpSession = avdumpList.sessions[avdumpList.sessionMap[fileId]];
 
-  const hash = useMemo(() => getEd2kLink(file), [file]);
+  const ed2kHashLink = useMemo(() => getEd2kLink(file), [file]);
```
hidden4003: We are using react compiler so this should be memoized automatically

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
```diff
@@ -84,11 +85,28 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
       } as const;
     }
 
+    const crc32 = file.Locations[0].RelativePath
```
hidden4003:
```suggestion
    const crc32 = file.Locations?.[0]?.RelativePath
```
Already used in the regex part but not here for some reason, also in the regex the `toUpperCase()` also needs `?.`

In general I would suggest extracting `file.Locations?.[0]?.RelativePath` into a variable.

hidden4003 (changes requested):

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
```diff
@@ -84,11 +85,28 @@ const AVDumpFileIcon = ({ file, truck = false }: { file: FileType, truck?: boole
...
+      && /\b[0-9A-F]{8}\b/i.exec(file.Locations[0].RelativePath)?.[0].toUpperCase();
```
hidden4003: Import regex from this file please https://github.com/ShokoAnime/Shoko-WebUI/blob/8480c8d0b4e1ddbc7e37401e235228015295e4ef/src/core/utilities/auto-match-regexes.ts#L11

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
pokyunn: This one i dont understand, just renamed var to make it explicit (as i have hash in arrow fun bellow)

harshithmohan: > We are using react compiler so this should be memoized automatically

It was already `useMemo`, only the var name changed to a better one

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
pokyunn: To be able to use it, I'll need to export this const (move outside try) and change from ?.[0] to ?.[1] as this one dont have boundaries \b so first match has the encapsulating () [] so on

pokyunn: And got cases where it matches "wrong" like (pattern has `|` )" so it always need double check 
image
so will need same `crc32Result[1] || crc32Result[2]`

Can I change the regex to mine? `/\b[0-9A-F]{8}\b/i` as it make always return at `[0]`
image

harshithmohan: You can update the regex in that file, it's not used anywhere else. As long as it works in intended, it's fine

harshithmohan: We actually discussed yesterday that it maybe could be
`/[\.(\[][0-9A-F]{8}[\.)\]]/i`

pokyunn: It can but will limit to 3 common variants (E943D68D) [E943D68D] .E943D68D.
If its ok, I agree

... and its a bit harder to read (for me)

pokyunn: > It can but will limit to 3 common variants (E943D68D) [E943D68D] .E943D68D. If its ok, I agree

And as it will need a nother pair of () to capture only it contents and not the bondaries like `/[\.(\[]([0-9A-F]{8})[\.)\]]/i`

pokyunn: Kept `/\b[0-9A-F]{8}\b/i` as the proposed will match to `.A1B2C3D4]` wich for me is inconsistency, but feel free to modify, only attendion if add capturegroup will need change `[0]` to `[1]`

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
hidden4003: My point was that we don't even need  `useMemo`, but if @harshithmohan thinks that we do need it then it can stay

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
harshithmohan: image

We don't care about filenames like this? :pensive:

pokyunn: 😔 In all these years in this vital industry, this is the first time i see it with _ around crc

pokyunn: > We don't care about filenames like this?  😔

But tbh @harshithmohan in this case is better just change to `/[^0-9A-Z]([0-9A-F]{8})[^0-9A-Z]/i` so only if it finds an 8digit hex, that does not have around any leter/number is valid
image

src/components/Utilities/Unrecognized/AvDumpFileIcon.tsx
harshithmohan: We don't need it, I'm just pointing out that it was pre-existing and not part of this PR

pokyunn: So removed.

hidden4003 (approved):

PR #1452, opened 2026-08-27
feat: dedicated server StatusPage for lifecycle states

harshithmohan:
## Summary

Adds a dedicated, full-bleed **StatusPage** (`/webui/status`) that is the single canonical surface for all non-ready server states — `Starting`, `Failed`, `Restarting`, `Shutting-down`, `Shutdown`, and `Offline` — plus an automatic recovery redirect back into the app once the server is `Started`. This fixes the auto-login gap that occurred when the server transitioned `Starting → Started` while the user was on the login page.

## Why

- `LoginPage` lived outside `AuthenticatedRoute`, so the `Starting` state was handled ad-hoc there and the `Starting → Started` transition was never reliably caught (auto-login gap during the boot window).
- A 502 during `Starting` produced a blank crossfade / stuck state because nothing canonical owned the "server not reachable yet" state.
- Auto-login could fire during the boot window, racing server readiness.
- Restart/shutdown had no canonical UX; SignalR toasts and route guards fought the in-progress action.

## Changes

- **New `serverLifecycle` slice** (`src/core/slices/serverLifecycle.ts`) — ephemeral UI intent (`idle` | `restarting` | `shutting-down`, `initiatedAt`). Deliberately **excluded from sessionStorage persistence** (both `loadState` and `saveState` in `src/core/localStorage.ts`) so a reload mid-restart never rehydrates a stale `restarting` and traps the user on a stuck StatusPage.
- **New `StatusPage.tsx`** (`src/pages/StatusPage.tsx`) — 7 variants (`loading`, `starting`, `failed`, `restarting`, `shutting-down`, `shutdown`, `offline`). Drive text via `variantDetailMap` and icons via `variantIconMap`. Polling at **2s / retry 5s** that backs off to **15s / 15s** once the server is unreachable. A **3-minute restart timeout** promotes `restarting → offline`; **2 consecutive failures** promotes `shutting-down → shutdown`. Immediate redirect on `Started` is guarded by `lastSeenStartedAtRef` + `sawDisruptionRef` + `STATUS_FRESH_MS` so the redirect only fires after a genuine disruption. Variants `failed`/`offline`/`shutdown` render action buttons — **"Restart Server"** (gated on `CanRestart`), **"Retry Now"**, and **"Retry Connection"** respectively — so recovery is possible directly from the page.
- **`AuthenticatedRoute` guard** (`src/core/router/AuthenticatedRoute.tsx`) — `isError` or active lifecycle `action` → `StatusPage`; `Waiting` → login; `Started` → authenticated children / login redirected; `Starting`/`Failed` → `StatusPage`; otherwise optimistically render when authenticated.
- **`/webui/status` route** added in `src/core/router/index.tsx`.
- **SignalR suppression** (`src/core/signalr/signalr.ts`) — `onreconnecting` / `onreconnected` / `onclose` and `requiresRestart` are gated on `serverLifecycle.action === 'idle'`, so StatusPage is the only UX during a user-initiated restart/shutdown.
- **`GeneralSettings` handoff** (`src/pages/settings/tabs/GeneralSettings.tsx`) — new Restart/Shutdown buttons (gated on `CanRestart`/`CanShutdown`) with `ConfirmationPromptModal`; `onConfirmRestart/onConfirmShutdown` dispatch `setAction` → mutate (`Init/Restart`, `Init/Shutdown`) → `navigate('/webui/status')`.
- **`LoginPage` redirect** (`src/pages/login/LoginPage.tsx`) — any non-`Started`/non-`Waiting` state (or `isError`) redirects to `/webui/status`; removed the bespoke `Starting`/`Failed` blocks from the login UI.
- **axios 401 fix** (`src/core/axios.ts`) — `fullPath` now uses a proper `/${url}` slash, adds `Settings` exclusion (`/api/v3/Settings`) for the boot window so settings reads don't trigger a logout.
- **Simplifications** — detail destructure, `!!` truthiness checks, retry-interval consolidation, icon map, etc.
- **Verified** — `pnpm lint` and `pnpm build` pass; dev override removed before commit; a11y/aria removed per codebase conventions.

## Testing

- `pnpm lint` and `pnpm build` pass.
- Manual restart / shutdown / starting flows verified end-to-end (StatusPage surfaces, redirects recover on `Started`).

### DIFF
--- a/src/pages/StatusPage.tsx
+++ b/src/pages/StatusPage.tsx
@@ -48,13 +48,13 @@ const variantDetailMap: Record<Variant, VariantDetail> = {
     button: 'Retry Now',
     headline: 'Shoko appears to be offline',
     title: 'Offline | Shoko',
-    whisper: 'You can also retry manually — automatic checks keep running either way.',
+    whisper: 'You can also retry manually. Automatic checks keep running either way.',
   },
   restarting: {
     headline: 'Shoko is restarting…',
     statusFallback: 'Waiting for the server to come back…',
     title: 'Restarting… | Shoko',
-    whisper: 'This can take a minute or two.',
+    whisper: 'This may take a moment.',
   },
   'shutting-down': {
     activity: 'Waiting for the server to stop…',
@@ -69,7 +69,7 @@ const variantDetailMap: Record<Variant, VariantDetail> = {
     title: 'Shut Down | Shoko',
   },
   starting: {
-    activity: 'This usually takes less than a minute.',
+    activity: 'This may take a moment.',
     headline: 'Shoko is starting up…',
     statusFallback: 'Preparing…',
     title: 'Starting… | Shoko',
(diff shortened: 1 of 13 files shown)


src/pages/StatusPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+  },
+  offline: {
+    activity: 'Retrying automatically…',
+    body: 'The server stopped responding. This usually means it crashed or the connection to it dropped.',
+    bodyTimeout:
+      'The server didn\'t come back within 3 minutes of restarting. It may still be starting, or something went wrong.',
+    button: 'Retry Now',
+    headline: 'Shoko appears to be offline',
+    title: 'Offline | Shoko',
+    whisper: 'You can also retry manually — automatic checks keep running either way.',
```
revam: Em dash?

src/pages/StatusPage.tsx
```diff
@@ -0,0 +1,353 @@
...
+  },
+  shutdown: {
+    body:
+      'The server process has stopped. To use Shoko again, start Shoko Server on the host machine, then retry the connection.',
+    button: 'Retry Connection',
+    headline: 'Shoko has shut down',
+    title: 'Shut Down | Shoko',
+  },
+  starting: {
+    activity: 'This usually takes less than a minute.',
```
revam: Instead of quoting a specific time frame for the "usual" duration, I'll recommend being a bit more abstract _somehow_. For instance, in my server **never** takes 'less than a minute' to boot, be it on my "prod" server or dev machine, both decent-ish hardware wise.

revam (commented):
Got asked to review text, so reviewed text. My human findings are below. Rest of 'em seem fine-ish.

hidden4003 (approved):

PR #1456, opened 2026-09-02
fix(auto-match): broaden trailing-native-title and related rules

Foowy:
## Targeted follow-up to #1450

Rescoped from the original broad-rule rewrite. Per review discussion, the WebUI auto-match rules are a **fallback path** (series select + auto-match episode in the edit-release-info modal) — the offline importer normally does the matching, and adding a WebUI regex per filename shape just grows maintenance. So this drops the speculative broadening (`marker-first`, `raws-3` tightening, scene-edition-tag / dot-underscore delimiter tolerance) and keeps only fixes that are bugs regardless of filename shape.

### Fixes in `src/core/utilities/auto-match-regexes.ts`

- **`reversed-1`**: the show-name class was `[^[]+]` — the trailing `]` is a required literal, so a plain `05 - Show Name [1080p].mkv` never matched and fell through to `fallback` as `05- Show Name`. Dropped the stray `]`.
- **`trash-anime`**: take the episode number from the `SxxExx` marker, not the trailing absolute number. The search and the cross-reference matcher both poll AniDB, which catalogues a multi-cour show as separate entries each numbered from 1, so `... - S02E08 - 021 - ...` is episode **8**, not 21. A single absolute-numbered entry keeps season 1 in Sonarr, so the two numbers are identical there. Also captures the season, and keeps the `SxxExx` range for multi-episode files.
- **`defaultTransform`**: theme songs matched by a rule with no `isThemeSong` group (`raws-*`, `trailing-native-title`) came out as `Episode`. `episodeType` is now set to `Credits` in the theme block where the theme song is actually detected — but only from the plain `Episode` default, so an `isSpecial`/`isTrailer` classification still wins.

### Also

- Documents the ordered / first-match-wins rule-set contract on `PathMatchRuleSet` (requested in review).
- Tests moved onto the Vitest suite added in #1459 (`tests/core/utilities/auto-match.test.ts`): updated `reversed-1` / `trash-anime` expectations and added multi-cour, multi-episode-range + theme-song-via-raws cases. New cases use generic placeholder filenames (`Generic Show Title`, `Show Name`) rather than real release names.

Verification: `pnpm test` (184 pass), `pnpm lint`.

### DIFF
--- a/src/core/utilities/auto-match-regexes.ts
+++ b/src/core/utilities/auto-match-regexes.ts
@@ -135,8 +135,12 @@ try {
       modifiedDetails.episodeStart = episode;
       modifiedDetails.episodeEnd = episode;
       // Rules that match theme songs via their own regex (raws-*, trailing-native-title) have no isThemeSong
-      // group, so detectEpisodeType never sees it - set the type here where the theme song is actually detected.
-      modifiedDetails.episodeType = 'Credits';
+      // group, so detectEpisodeType never sees it - set the type here. Only from the plain 'Episode' default,
+      // so a more specific classification (Special from an isSpecial group, etc.) still wins, matching the
+      // isSpecial > isThemeSong precedence in detectEpisodeType.
+      if (modifiedDetails.episodeType === 'Episode') {
+        modifiedDetails.episodeType = 'Credits';
+      }
     }
     if (modifiedDetails.episodeStart === 0) {
       const trailerCheckResult = TrailerCheckRegex.exec(originalDetails.filePath);
@@ -215,10 +219,12 @@ try {
       // Take the episode number from the SxxExx marker, not the trailing absolute number: the search and the
       // cross-reference matcher both poll AniDB, which catalogues a multi-cour show as separate entries each
       // numbered from 1, so "... - S02E08 - 021 - ..." is episode 8, not 21. A single absolute-numbered entry
-      // keeps season 1 in Sonarr, so the two numbers are identical there.
+      // keeps season 1 in Sonarr, so the two numbers are identical there. The E-range stays inside `episode`
+      // (defaultTransform splits it), and `year` is a bare \d{4} - not (?:19|20)\d{2} - because Sonarr can emit
+      // any 4-digit year here.
       regex:
         // oxlint-disable-next-line no-useless-escape
-        /^(?<showName>.+?(?: \((?<year>\d{4})\))) - (?:(?<isSpecial>S00?)|S(?<season>\d+))E(?<episode>\d+)(?:-E?\d+)? - \d+(?:-\d+)? - (?<episodeName>.+?(?=\[)).*?(?:-(?<releaseGroup>[^\[\] ]+))?\s*\.(?<extension>[a-zA-Z0-9_\-+]+)$/id,
+        /^(?<showName>.+?(?: \((?<year>\d{4})\))) - (?:(?<isSpecial>S00?)|S(?<season>\d+))E(?<episode>\d+(?:-E?\d+)?) - \d+(?:-\d+)? - (?<episodeName>.+?(?=\[)).*?(?:-(?<releaseGroup>[^\[\] ]+))?\s*\.(?<extension>[a-zA-Z0-9_\-+]+)$/id,
       transform: defaultTransform,
     },
     {
--- a/src/core/utilities/auto-match-regexes.ts
+++ b/src/core/utilities/auto-match-regexes.ts
@@ -134,6 +134,9 @@ try {
 
       modifiedDetails.episodeStart = episode;
       modifiedDetails.episodeEnd = episode;
+      // Rules that match theme songs via their own regex (raws-*, trailing-native-title) have no isThemeSong
+      // group, so detectEpisodeType never sees it - set the type here where the theme song is actually detected.
+      modifiedDetails.episodeType = 'Credits';
     }
     if (modifiedDetails.episodeStart === 0) {
       const trailerCheckResult = TrailerCheckRegex.exec(originalDetails.filePath);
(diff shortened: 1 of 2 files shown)


Foowy:
@harshithmohan tagging you for review since you handled the #1450 review. This is the follow-up covering the filename shapes that slipped past it. One deliberate behaviour change flagged in the description (`trash-anime` dropping `(year)` from the seed)

Foowy:
Pushed a second commit (`7931bc4e`): `trash-anime` now takes the episode number from `SxxExx` instead of the trailing `- <absolute> -` segment. Rationale: the search and cross-ref matching poll AniDB, which splits a multi-cour show into separate entries numbered from 1, so `... - S02E08 - 021 - ...` should match episode 8, not 21. For a single absolute-numbered AniDB entry Sonarr keeps season 1, so nothing changes there. It also happens to fix specials (`S00E01 - 12` -> special 1). Easy to drop if you'd rather keep the absolute behaviour.

Foowy:
Responses folded into the PR description under **Review responses**. Summary:

- **Specials to AniDB special number:** correct as-is. A special's `epno` is a per-type 1-based sequence (`S1`, `S2`, ...) unrelated to the absolute count ([UDP API def](https://wiki.anidb.net/UDP_API_Definition), episode `type` 2), so the trailing absolute `12` was never the special number and `S00E01` -> special 1 is right.
- **`marker-first` bare-title seed when a year is present:** kept as-is. The year has nowhere to go (`PathDetails` has no year field) and the ` S{season}` drop comes from the pre-existing shared `!year` guard, not this PR. Prior output was the unusable `S03E`. A disambiguator would be a separate change to the shared guard.
- **`marker-first` not tagging `S00` as special:** fixed in `9ff37529`.
- **`\d{4}` vs `(?:19|20)\d{2}`:** intentional, now noted in the `trash-anime` comment.
- **`raws-3` pre-existing season bug:** documented in the `raws-3` bullet.

harshithmohan:
I have a bigger question, how is this even supposed to work? For example, if you have Dr. Stone S04E37, it maps to nothing. Wouldn't this only work for anime which have a single season? And I don't know if I trust that numbering fully either.
@hidden4003 thoughts?

Foowy:
The number isn't looked up against AniDB directly. In the Link Files flow the user first searches and picks an AniDB entry, and then the detected episode number is matched against *that entry's* own episode list (`LinkFilesTab.tsx` ~L340: `find(episodes, e => e.Type === type && e.EpisodeNumber === episodeStart)`). Every AniDB entry starts counting from episode 1. If the number doesn't match anything in the picked entry, that file is just skipped and you link it by hand, so a wrong guess can't quietly create a wrong link.

Because the match is against a single AniDB entry that counts from 1, the number we want is the episode's position inside its own cour/part:

- Sequel season, one cour, e.g. `Dr. Stone Stone Wars - S02E05 - 029`: AniDB lists "Stone Wars" as its own entry, and this is episode 5 of it. Season-relative gives `5` and matches. The old absolute `029` matched nothing. AniDB splits almost every multi-season show like this, so this is the common case and the one the commit is for.
- Single season, e.g. `S01E08 - 008`: `8` either way.
- A TVDB season that lumps several AniDB entries together (your `Dr. Stone S04` case, if S04 is more than one cour): season-relative `37` matches nothing, and the old absolute matched nothing either. Nothing parsed from the filename can fix this one, it needs the TVDB/TMDB to AniDB mapping data (refactor/far larger change for suspected minimal gain). So it's no worse than today, just still skipped. 

One option that sidesteps the "which number" question entirely: have the parser return both the season-relative and the absolute number, and let the auto-fill try season-relative first, then fall back to absolute if that doesn't match anything in the selected entry. No mapping data or network lookups, just a bit more logic in detectShow plus the three spots that consume it. It won't fix the lumped-together-season case, but it removes the guesswork and the worst outcome is still just "skipped, link manually", same as today.

So "only works for single-season anime" is actually the old behavior. Absolute numbering only lined up when the whole show was one AniDB entry. The change makes sequel seasons work too, without breaking the lumped-together case.

Can drop the second commit for now and revisit the episode-numbering in a separate discussion if you'd prefer.

Further items:

The heavier alternative would be doing it properly on the Server side: a new endpoint that takes the parsed title plus SxxExx, looks the show up on TMDB, resolves the actual episode there (TMDB has "episode groups" that model absolute and alternate orderings), then maps that to an AniDB episode through the TMDB/AniDB cross-references Shoko already keeps. The snag is that for a file that isn't recognised yet there's no AniDB link, so there's no cross-reference to follow, and matching a TMDB show to the right AniDB entry is its own problem. It would mean new Server logic, a new API endpoint, and WebUI wiring, just noting where a real fix would live.

hidden4003:
>I have a bigger question, how is this even supposed to work? For example, if you have Dr. Stone S04E37, it maps to nothing. Wouldn't this only work for anime which have a single season? And I don't know if I trust that numbering fully either.

All anime on AniDB has only one season, so we are talking here unrecognised files utility is where these are used? As far as I remember there we use anidb shows, do we actually use tmdb season data in there?

Also at this point this file needs to have an extensive test suite with all the supported filenames so that we can see regressions when they occur. 

What is that table even, what does seeded before and seeded after mean?

Overall that whole file needs some documenting on which regexes are used where so that they are not misused in different parts of the code.

Foowy:
> > I have a bigger question, how is this even supposed to work? For example, if you have Dr. Stone S04E37, it maps to nothing. Wouldn't this only work for anime which have a single season? And I don't know if I trust that numbering fully either.
> 
> All anime on AniDB has only one season, so we are talking here unrecognised files utility is where these are used? As far as I remember there we use anidb shows, do we actually use tmdb season data in there?
> 
> Also at this point this file needs to have an extensive test suite with all the supported filenames so that we can see regressions when they occur.
> 
> What is that table even, what does seeded before and seeded after mean?
> 
> Overall that whole file needs some documenting on which regexes are used where so that they are not misused in different parts of the code.

"seeded before / after": those columns are the show name the tool guesses from the filename and drops into the search box for you when you open the unrecognised-files / Link Files screen. It's whatever detectShow pulls out as showName. You can edit it before hitting search. "before" is how master guesses it today, "after" is with this PR. The Episode column is the episode number it guesses; once you pick an AniDB series, the auto-linker looks for that number in that series' episode list. I'll add this as a legend above the table.

Test suite: agree it needs one. The repo has no test runner today, so that's a small setup step first.

Per-regex docs: fair, I can add a comment block mapping each rule to where its output is consumed.

Foowy:
Pausing work, until final decision of regex ingestion path is settled.

PR #1459, opened 2026-09-06
feat(test): add Vitest unit tests for filterTree and auto-match logic

harshithmohan:
## Summary

Adds the repo's first unit-test layer (Vitest 5) for regression protection of the two highest-risk pure modules, per RFC [#1458](https://github.com/ShokoAnime/Shoko-WebUI/discussions/1458).

**Constraint:** Tests are deliberately limited to high-risk modules only — no coverage tooling, no DOM/UI assertions.

## Changes

- **New devDependency:** `vitest@5.0.0`
- **New `vitest.config.ts`:** Standalone config (does NOT load `vite.config.mjs`); alias `@` → `src/`, `environment: 'node'`, `include: ['tests/**/*.test.ts']`, `globals: false`
- **`package.json`:** New scripts `test` (`vitest run`) and `test:watch`
- **`tsconfig.json`:** `rootDir` removed; `include` adds `./tests/`
- **`.oxlintrc.json`:** `ignorePatterns` adds `!tests/`, `!tests/**`
- **`.dprint.json`:** `includes` adds `tests/**`
- **`.github/workflows/Lint-PR.yml`:** New `Test` step (`pnpm test`) after Lint
- **New tests:**
  - `tests/core/utilities/filterTree.test.ts` — ~110 tests covering `parseFilterTree`/`buildFilterTree` round-trips, folding, negation, tree surgery; fixtures mirror real Shoko Server wire data incl. production FilterPreset trees
  - `tests/core/utilities/auto-match.test.ts` — ~68 tests covering all 13 `PathMatchRuleSet` rules, transform behavior via public `detectShow`
- **`AGENTS.md`:** Updated to document the test layer and CI gate

## Suite

2 files / 183 tests, ~250ms; `pnpm lint` + `pnpm tscheck` green.

Relates to #1458


hidden4003 (approved):
## Review: PR #1459 + RFC #1458 (updated)

### RFC #1458

The author edited the RFC to mark all three open questions **resolved per hidden4003's reply**: CI extends `Lint-PR.yml` (no separate workflow, no explicit install step since `pnpm/setup@v1` already installs), determinism uses structural assertion (no `crypto.randomUUID` stubbing), and the `useEffectEvent`/Phase 3 assertion target is confirmed. An inline note was also added to §2C referencing the resolution. Correctly incorporated — nothing left unresolved.

### PR #1459

Implements Phase 0–2 of the RFC exactly as scoped. The branch was checked out, dependencies installed, and everything run directly rather than just reading the diff:

- **`pnpm test`** — 183/183 pass (2 test files, ~565ms)
- **`pnpm lint`** (dprint + oxlint + stylelint) — clean
- **`pnpm tscheck`** — clean
- **Config changes** (`vitest.config.ts`, `tsconfig.json` `rootDir` removal, `.oxlintrc.json`/`.dprint.json` scope extensions, `Lint-PR.yml` new `Test` step, `AGENTS.md` updates) all match what the RFC specified, including the exact CI-placement and determinism decisions from the resolved open questions.

**Test quality is genuinely strong, not just green.** The auto-match suite documents actual implementation quirks/discrepancies inline as comments (e.g. `raws-1`'s regex captures a `source` group that `detectShow` silently drops — the test asserts `source: null` with a `// DISCREPANCY` comment rather than hiding it), and the `filterTree` suite round-trips against real production `FilterPreset` fixtures (Favorites, Continue Watching, Newly Added Series with its full function-tree preservation), which is exactly the "ground truth, not aspirational" bar the RFC set. No skipped tests, no placeholder assertions.

**Needs a fix before/shortly after merge:** `pnpm test` prints a Vite warning that `vitest.config.ts` uses ESM syntax in a file loaded as CommonJS (`package.json` has no `"type": "module"`), which Vite says becomes a hard requirement in a future major version. This is the same class of issue `vite.config.mjs` hit and was fixed for in `f8edee10` ("Fix vite config", Aug 16 2026) — that fix relied on the `.mjs` extension forcing native-ESM loading regardless of `package.json`'s `"type"` field, plus `with { type: 'json' }` / `import.meta.dirname` adjustments. `vitest.config.ts` doesn't get that same treatment from its extension, so the warning fires. Recommend renaming to `vitest.config.mts` (or adding an equivalent fix) so it doesn't become a breaking upgrade blocker later, the way the main config's issue did.

**Overall:** ready to merge — implementation faithfully matches the agreed-upon RFC scope, all gates are green, and the test content itself is high-signal rather than rubber-stamped coverage. The config-loader warning is a minor, pre-existing-pattern fix, not a blocker.

Issue #922, opened 2023-01-29
feature: renaming episodes


papaj-na-wrotkach:
Add ability for user to manually rename episodes (just like for series)

revam:
I did a thing in the spur of the moment, so now the API supports it, but we still need to add UI for it, so I'm transferring this issue to the Web UI repository for now.

ElementalCrisis:
We currently do not plan to support this feature due to its limited use case, as our resources can be better utilized on more pressing tasks.

While we won't prevent anyone from attempting to implement this feature, please be aware that it will not be a straightforward pull request. You'll need to get the design approved by us first.

PR #1467, opened 2026-09-14
feat(episode): add ability to override episode title

hidden4003:
## Summary
Adds a UI for manually overriding an episode's title, closing #922. A pencil icon next to the episode title opens a modal where you can type a custom title or pick from AniDB's alternate titles, with a reset option to clear the override. This is backed by the existing `Episode/{id}/OverrideTitle` server endpoint, mirroring the series title-override pattern already in the app.


src/components/Collection/Episode/EditEpisodeTitleModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const handleReset = () => {
+    if (isPending) return;
+    overrideTitle('', { onSuccess: onRequestClose });
+  };
+
+  useToggleModalKeybinds(show, 'modal');
+  useToggleModalKeybinds(!show, 'primary');
+  useHotkeys('escape', onRequestClose, { scopes: 'modal' });
```
harshithmohan: ### `Escape` closes the modal while the save is pending

The Escape hotkey is bound unconditionally, so it defeats the intent of `onRequestClose={isPending ? undefined : onRequestClose}` (which only guards the backdrop click). During an in-flight save, Esc still closes the modal:

```tsx
useHotkeys('escape', () => !isPending && onRequestClose(), { scopes: 'modal' });
```

src/components/Collection/Episode/EditEpisodeTitleModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  useHotkeys('enter', handleSave, { scopes: 'modal' });
```
harshithmohan: ### `Enter`-to-save doesn't fire while the input is focused

`react-hotkeys-hook` ignores key events whose target is a form tag (`input`/`textarea`/`select`) unless `enableOnFormTags` is set. Since the title `<Input autoFocus>` is focused as soon as the modal opens — the normal case — this hotkey never runs, so the Enter shortcut is effectively dead. Either add `enableOnFormTags: true` or drop the Enter binding:

```tsx
useHotkeys('enter', handleSave, { scopes: 'modal', enableOnFormTags: true });
```

harshithmohan (changes requested):
The modal and mutation follow the existing series-title-override pattern nicely, and the endpoint contract checks out. A few changes requested below.

src/components/Collection/Episode/EditEpisodeTitleModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [title, setTitle] = useState('');
+
+  const episodeId = episode.IDs.ID;
+  const { isPending, mutate: overrideTitle } = useOverrideEpisodeTitleMutation(episodeId, seriesId, nextUp);
+
+  useEffect(() => {
+    if (show) setTitle(episode.Name);
+  }, [episode.Name, show]);
+
+  const handleSave = () => {
```
harshithmohan, line 31: ### Add success and error toast feedback at the call site

`useOverrideEpisodeTitleMutation` has no error handling, and mutations in `src/core/react-query/episode/mutations.ts` are intentionally toast-free (`useWatchEpisodeMutation`/`useHideEpisodeMutation`), with their consumers handling feedback at the call site. There's also no global error toast in the axios interceptors, so a failed save — e.g. a 403 for non-admin users, since the endpoint is `[Authorize("admin")]` — fails silently and leaves the modal open with no indication. Toasts belong here, alongside the existing `onSuccess: onRequestClose`:

```tsx
import toast from '@/core/toast';

const handleSave = () => {
  if (isPending) return;
  overrideTitle(title, {
    onSuccess: () => {
      toast.success('Episode title updated successfully!');
      onRequestClose();
    },
    onError: () => toast.error('Episode title could not be updated!'),
  });
};

const handleReset = () => {
  if (isPending) return;
  overrideTitle('', {
    onSuccess: () => {
      toast.success('Episode title reset to default!');
      onRequestClose();
    },
    onError: () => toast.error('Episode title could not be reset!'),
  });
};
```

harshithmohan (approved):

